PR #10, opened 2025-03-20
feat: ✨ initial slasher implementation

TDemeco:
This PR:
- Sets up the slasher infrastructure with the base functionality required (in `ISlasher.sol`, `SlasherBase.sol` and `SlasherBaseStorage.sol`) and adds the tests for it (in `SlasherBase.t.sol`).
- Adds an implementation of a more complex slasher (in `IVetoableSlasher.sol` and `VetoableSlasher.sol`) and tests for it (in `VetoableSlasher.t.sol`).
- Modifies the `ServiceManagerBase` contract to use the new `VetoableSlasher` contract to manage slashing.
- Updates mocks and tests to reflect the newly added functionality.


contracts/src/interfaces/IVetoableSlasher.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import {IAllocationManager} from
+    "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
+import {ISlasher} from "./ISlasher.sol";
+import {IServiceManager} from "./IServiceManager.sol";
```
gonzamontiel: This interface is not used in the file at least.

contracts/src/interfaces/IVetoableSlasher.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: BUSL-1.1
```
gonzamontiel, line 1: Why are we adding an interface to the repo that is part of `eigenlayer-middleware`, can't we just use it _as-is_ ?

contracts/src/interfaces/ISlasher.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BUSL-1.1
```
gonzamontiel, line 1: Also with this one, do we really need to copy the same file as in `eigelayer-middleware` ?

contracts/src/middleware/SlasherBase.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    /// @notice Internal function to execute a slashing request
+    /// @param _requestId The ID of the slashing request to fulfill
+    /// @param _params Parameters defining the slashing request including operator, strategies, and amounts
+    /// @dev Calls AllocationManager.slashOperator to perform the actual slashing
+    function _fulfillSlashingRequest(
+        uint256 _requestId,
+        IAllocationManager.SlashingParams memory _params
+    ) internal virtual {
+        allocationManager.slashOperator({avs: serviceManager.avs(), params: _params});
```
gonzamontiel, line 37: I see that here we are replacing the `slashingRegistryCoordinator` by the `serviceManager`. So our DataHaven service manager contract is the only one that for now can slash, right?

contracts/src/interfaces/IVetoableSlasher.sol
TDemeco: Good catch! It was leftover from a previous iteration of the interface, when I was testing things out. I'll remove it.

contracts/src/interfaces/IVetoableSlasher.sol
TDemeco: This might answer most of your questions:

As you might have seen, we do not have the `eigenlayer-middleware` repo as a dependency at all. That's because that repo is currently in a 'limbo' state because of the upgrade from V1 to V2, so some things are updated, some are not, and not everything works as it should. That's why initially @ffarall set up the `ServiceManager` to not required that repo and why I'm following his footsteps with the `Slashers`.

Regarding specifically the `VetoableSlasher`, what I did was adapt it to work with our structure but as you might suspect it is not the final `Slasher` that DataHaven is going to use. The purpose it mainly has right now is both to allow more comprehensive testing of the `ServiceManager` (since we couldn't test the slashing functionality before) and also as boilerplate for when we define, design and implement the actual slasher, which will probably be more complex.

contracts/src/interfaces/ISlasher.sol
TDemeco: Same answer as above^

contracts/src/middleware/SlasherBase.sol
TDemeco: My ideal situation would be that the `ServiceManager` would have a `queueUpSlashingRequest` function that receives both who has to be slashed and a proof that that Operator misbehaved and internally calculates how much to slash the Operator for (according to previous performance for example). This would allow anyone to punish misbehaving Operators through the `ServiceManager`, so the `ServiceManager` should be the only one allowed to actually queue up a slashing request in the `Slasher` that we end up using.

gonzamontiel: I agree, ideally I would encapsulate the slashing conditions in a different contract like SlashingCoordinator or so on, but with a similar approach to what you are describing here,

contracts/src/interfaces/ISlasher.sol
```diff
@@ -0,0 +1,39 @@
...
+        uint256[] wadsToSlash,
+        string description
+    );
+}
+
+/// @title ISlasher
+/// @notice Base interface containing shared functionality for all slasher implementations
+interface ISlasher is ISlasherErrors, ISlasherEvents {
+    /// @notice Returns the address authorized to create and fulfill slashing requests
+    function slasher() external view returns (address);
```
ffarall, line 35:
```suggestion
    /// @notice Returns the address authorized to create and fulfil slashing requests
```

contracts/src/interfaces/IVetoableSlasher.sol
```diff
@@ -0,0 +1,86 @@
...
+    error SlashingRequestIsCancelled();
+    /// @notice Thrown when attempting to modify a slashing request that does not exist
+    error SlashingRequestNotRequested();
+}
+
+interface IVetoableSlasherTypes {
+    /// @notice Represents the status of a slashing request
+    enum SlashingStatus {
+        Requested,
+        Cancelled,
```
ffarall: Why having a `Cancelled` status instead of deleting it?

contracts/src/interfaces/IVetoableSlasher.sol
```diff
@@ -0,0 +1,86 @@
...
+    /// @notice Cancels a pending slashing request
+    /// @param requestId The ID of the slashing request to cancel
+    /// @dev Can only be called by the veto committee during the veto period
+    function cancelSlashingRequest(
+        uint256 requestId
+    ) external;
+
+    /// @notice Executes a slashing request after the veto period has passed
+    /// @param requestId The ID of the slashing request to fulfill
+    /// @dev Can only be called by the authorized slasher after the veto period
```
ffarall, line 82:
```suggestion
    /// @param requestId The ID of the slashing request to fulfil
```

contracts/src/interfaces/IVetoableSlasher.sol
```diff
@@ -0,0 +1,86 @@
...
+    function fulfillSlashingRequest(
```
ffarall: To be fair, "fulfill" is correct in American English. If you want to keep it like this, it's fine, if you change it to the British spelling (fulfil) make sure to change it everywhere else in the project.

```suggestion
    function fulfilSlashingRequest(
```

contracts/src/middleware/SlasherBase.sol
```diff
@@ -0,0 +1,61 @@
...
+
+/// @title SlasherBase
+/// @notice Base contract for implementing slashing functionality in an EigenLayer AVS
+/// @dev Provides core slashing functionality and interfaces with EigenLayer's AllocationManager
+abstract contract SlasherBase is SlasherStorage {
+    /// @notice Ensures only the authorized slasher can call certain functions
+    modifier onlySlasher() {
+        _checkSlasher(msg.sender);
+        _;
+    }
```
ffarall, line 19: I see this is used for `queueSlashingRequest`, but why would we have an account being the only one who can queue a slashing request?

This is just to understand better myself. One good answer could be "the registered Slasher would be the Gateway contract which validates slashing messages coming from our Substrate chain, and dispatches them as slashing requests to this contract".

Let me know if that's what you had in mind, or something like that.

ffarall: Looking at your response on another comment, I see that the intended slasher is the Service Manager.

contracts/src/middleware/VetoableSlasher.sol
```diff
@@ -0,0 +1,121 @@
...
+            VetoPeriodPassed()
+        );
+        require(
+            slashingRequests[requestId].status == IVetoableSlasherTypes.SlashingStatus.Requested,
+            SlashingRequestNotRequested()
+        );
+
+        slashingRequests[requestId].status = IVetoableSlasherTypes.SlashingStatus.Cancelled;
+        emit SlashingRequestCancelled(requestId);
+    }
```
ffarall: Have we considered a way to free up cancelled or completed slashing requests?

contracts/src/middleware/ServiceManagerBase.sol
```diff
@@ -97,12 +108,37 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage, IAVSRegistrar
...
-     * Forwards the call to the AllocationManager.slashOperator() function
+     * Forwards the slashing request to the set slasher
+     * @dev This function is now an alias for queueSlashingRequest for compatibility
      */
     function slashOperator(
         IAllocationManager.SlashingParams calldata params
     ) external virtual onlyOwner {
-        _allocationManager.slashOperator(address(this), params);
+        this.queueSlashingRequest(params);
+    }
```
ffarall: Compatibility with what? Why not deleting it? I don't see it being part of an interface that we have to abide by, for instance.

contracts/src/middleware/ServiceManagerBase.sol
```diff
@@ -97,12 +108,37 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage, IAVSRegistrar
...
+     * Queue a slashing request in the vetoable slasher
+     * @param params Parameters defining the slashing request
+     * @dev Can only be called by the owner
+     */
+    function queueSlashingRequest(
+        IAllocationManager.SlashingParams calldata params
+    ) external virtual onlyOwner {
+        require(address(_slasher) != address(0), "Slasher not set");
+        _slasher.queueSlashingRequest(params);
+    }
```
ffarall: In our AVS contract this wouldn't be `onlyOwner` but we would have the verification implemented to check the slashing, right? Which would be checking for the slashing message in a finalised Beefy block.

ffarall: Btw, now that I think of it, this `ServiceManagerBase` doesn't need to be an `abstract` contract, because we provide these "permissioned" implementations.

contracts/test/SlasherBase.t.sol
```diff
@@ -0,0 +1,261 @@
...
+        // Do nothing, just for testing the modifier
+    }
+
+    // Expose the internal _checkSlasher function for testing
+    function checkSlasher(
+        address account
+    ) external view {
+        _checkSlasher(account);
+    }
+}
```
ffarall: Shouldn't this be in our `mocks` folder?

contracts/test/SlasherBase.t.sol
```diff
@@ -0,0 +1,261 @@
...
+
+    function setUp() public virtual {
+        _deployMockEigenLayerAndAVS();
+        _setUpDefaultStrategiesAndMultipliers();
+
+        // Set up roles for testing
+        nonServiceManagerRole = address(0x5678);
+
+        // Deploy the SlasherMock contract
+        slasherContract = new SlasherMock(allocationManager, serviceManager);
```
ffarall: I understand you deploy this to test some functionalities of the `BaseSlasher`. But it's confusing, because in `_deployMockEigenLayerAndAVS()` you also deploy a `VetoableSlasher`, so it is not clear at a glance why you're deploying a slasher again.

contracts/test/VetoableSlasher.t.sol
```diff
@@ -0,0 +1,372 @@
...
+        _setUpDefaultStrategiesAndMultipliers();
+
+        // Set up roles for testing
+        nonServiceManagerRole = address(0x5678);
+        vetoCommittee = address(0xabcd);
+        nonVetoCommittee = address(0xdcba);
+
+        // Deploy the VetoableSlasher contract
+        vetoableSlasher =
+            new VetoableSlasher(allocationManager, serviceManager, vetoCommittee, vetoWindowBlocks);
```
ffarall: But haven't you deployed a `VetoableSlasher` already in `_deployMockEigenLayerAndAVS`?

contracts/test/VetoableSlasher.t.sol
```diff
@@ -0,0 +1,372 @@
...
+        uint256 requestId = vetoableSlasher.nextRequestId();
+
+        vm.prank(address(serviceManager));
+        vetoableSlasher.queueSlashingRequest(params);
+
+        return requestId;
+    }
+
+    // Helper function to extract SlashingRequest from storage
+    function getSlashingRequest(
```
ffarall:
```suggestion
    function _getSlashingRequest(
```

contracts/src/interfaces/IVetoableSlasher.sol
TDemeco: The idea is to be able to audit on-chain both the validators slashings and the committee actions without needing an indexer for the events. Since slashing should be a rare occurrence, and the veto committee canceling a slash should be even rarer, this should not be a big burden on the on-chain storage. I wouldn't mind just deleting it from storage though if you prefer

contracts/src/middleware/SlasherBase.sol
TDemeco: Yes! The Service Manager is the one which currently has the logic to validate and queue up a slashing request, although we can easily change this if we end up with a different design for contract interactions (for example as you said, directly processing slashing messages from the bridge and queueing up slashing requests that way).

contracts/src/middleware/VetoableSlasher.sol
TDemeco: Following up my answer above about why the slashing requests are not currently being deleted, regarding this we could use the `delete` keyword since we have the `requestId`

contracts/src/middleware/ServiceManagerBase.sol
TDemeco: In regards to the first point, that's exactly the design I was thinking about. Right now, only the owner of the AVS can queue up a slashing request but that's for simplification since we haven't delved into the slashing process in depth yet. The flow I thought about was:
1. A slashing has to occur this epoch since a validator (or multiple validators) failed to produce a block for a slot that it was selected. Slashing metadata (who to slash and for how much) is accumulated until the end of the epoch in a Merkle tree.
2. A slashing message is added to the BEEFY `extra` field of the final block of the epoch. This message contains the root of the slashing tree.
3. Once the BEEFY block gets relayed, anyone can provide a proof that a certain validator has to be slashed, in a similar fashion in how Operators can provide proofs to receive their rewards. This has to be incentivized in some way.
4. Any requests to withdraw stake from validators have to wait two epochs: one to deregister them as validators in Substrate (or update their weight if only withdrawing a partial amount) and one to make sure they didn't misbehave in their last epoch

contracts/test/VetoableSlasher.t.sol
TDemeco: Yeah, I first wrote this tests and then I added the slasher deployment to the mock deployment and forgot to update it here. Fixed

contracts/test/SlasherBase.t.sol
TDemeco: Updated the comment to make it more clear why we are doing this

contracts/test/SlasherBase.t.sol
TDemeco: Yep! I forgot to move it to a separate file. Updated

contracts/src/middleware/ServiceManagerBase.sol
TDemeco: You're right, when I changed this it seems I forgot that we are not inheriting anything from `eigenlayer-middleware` 😬

contracts/src/interfaces/IVetoableSlasher.sol
gonzamontiel: I think the questions here are: 
1) Do we want to provide a historic record of Cancelled ones "easily" ? If the answer is yes, then
 2) are we still indexing the events for other reasons? if the answer is yes, I would delete it, because we would already have a way of providing this info. 

In the other hand, if we are in doubt, I would keep it because until we are really live we have some time, and we can always upgrade and change the storage.

ffarall: In principle, we're not indexing them. The optimiser in me wants to clean up storage. The relaxed me wants to say just leave it. I side with the optimiser this time, but I can live with it if you keep it as it is.

contracts/src/middleware/VetoableSlasher.sol
TDemeco: Changed it so we free up storage, tracking of events should be done if a slashing history is required

contracts/src/interfaces/IVetoableSlasher.sol
TDemeco: Changed it so we free up the storage when a slashing request gets cancelled or fulfilled. Let's remember to add tracking slashing events if we ever need a history of them

PR #12, opened 2025-03-24
misc: add rust audit action to the workflow

undercover-cactus:
This PR add a github action that check for reported vulnerabilities in dependencies. It can be triggered by two ways : 
* Change in Cargo.toml or Crago.lock
* Once a week

Because it checks dependencies of dependencies it should not be required to pass a PR.


.github/workflows/rust-audit.yml
```diff
@@ -0,0 +1,24 @@
+# Rust Audit: CI for Rust audit of the operator node
+#
+# Overview:
+# 1. Check all the dependencies from Cargo.lock for reported vulnerabilities
+
+name: Security audit
```
ffarall: This sounds way more severe that it should 😂. If this fails, I'd be concerned. Maybe:
```suggestion
name: Audit Rust dependencies
```

gonzamontiel: I'd go even more clearer, like "Check dependencies vulnerabilities" or "Check Cargo vulnerabilities", we are not particularly auditing the code but just checking "Cargo audit", aren't we?

PR #13, opened 2025-03-24
feat(operator): 🏗️  add frontier client 

gonzamontiel:
- [x] Add frontier client dependencies
- [x] Add ethereum client config
- [x] Add changes to cli.rs
- [x] Add changes to service.rs
- [x] Add runtime apis
- [x] Add changes to chain spec
- [ ] Add tests

### DIFF
--- a/operator/Cargo.toml
+++ b/operator/Cargo.toml
@@ -15,7 +15,6 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-
 # Local
 datahaven-runtime = { path = "./runtime", default-features = false }
 datahaven-runtime-common = { path = "./runtime/common", default-features = false }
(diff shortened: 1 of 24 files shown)


operator/runtime/Cargo.toml
```diff
@@ -107,20 +108,24 @@ std = [
 	"pallet-offences/std",
 	"pallet-preimage/std",
 	"pallet-scheduler/std",
+    "pallet-babe/std",
+    "pallet-beefy/std",
+    "pallet-beefy-mmr/std",
+    "pallet-balances/std",
+    "pallet-grandpa/std",
+    "pallet-mmr/std",
+    "pallet-session/std",
```
stiiifff: These look like duplicates declarations.

gonzamontiel: good catch!

operator/Cargo.toml
```diff
@@ -24,24 +24,27 @@ pallet-validator-set = { path = "./pallets/validator-set", default-features = fa
...
 hex-literal = { version = "0.3.4" }
-codec = { version = "3.6.12", default-features = false, package = "parity-scale-codec" }
+codec = { package = "parity-scale-codec", version = "3.6.12", default-features = false, features = ["derive"] }
 blake2-rfc = { version = "0.2.18", default-features = false }
 impl-serde = { version = "0.5.0", default-features = false }
 libsecp256k1 = { version = "0.7", default-features = false }
 sha3 = { version = "0.10", default-features = false }
 hex = { version = "0.4.3", default-features = false }
+flume = "0.10.9"
+url = "2.2.2"
```
ffarall: Not necessarily to do now, but we'd like to keep the crates here in alphabetical order to find them easier later.

operator/runtime/src/apis.rs
```diff
@@ -483,4 +566,252 @@ impl_runtime_apis! {
...
+                // chain_id 8 bytes
+                // nonce: 32
+                // max_priority_fee_per_gas: 32
+                // max_fee_per_gas: 32
+                // gas_limit: 32
+                // action: 21 (enum varianrt + call address)
+                // value: 32
+                // access_list: 1 (empty vec size)
+                // 65 bytes signature
+                258;
```
ffarall, line 636: What's up with this?

operator/pallets/validator-set/src/lib.rs
```diff
@@ -35,6 +35,8 @@
 //! and [`pallet_session::Config::ValidatorIdOf`] must be [`ConvertInto`].
 
 #![cfg_attr(not(feature = "std"), no_std)]
+// We need this because it clashes with Polkadot macro pallet::pallet
+#![allow(clippy::manual_inspect)]
```
ffarall, line 39: Just out of curiosity, does it throw an error or a warning otherwise?

operator/Cargo.toml
gonzamontiel: great idea, I've created a PR https://github.com/Moonsong-Labs/datahaven/pull/33

operator/pallets/validator-set/src/lib.rs
gonzamontiel: It's a warning, I dind't check if our CI runs clippy as pedantic (warnings are errors) but at least that was the case in Moonbeam, so I try to fix or mark explicitly the warnings.

operator/runtime/src/apis.rs
gonzamontiel: I didn't dig a lot into this, it's a calculation made by Moonbeam, it could be that we need to mimic it or to change it to something else, I still don't have a lot of context. @ahmadkaouk  wdyt ?

ahmadkaouk: In moonbeam the transaction length is accounted for in the `proof_size_base_cost` since the call within the extrinsic is part of the Proof of Validity (PoV). However, since Datahaven is  a solo chain, this is not relevant since we're not bound by the proof size limitations that apply to parachains like Moonbeam. We should consider removing the proof_size constraints for Datahaven.

operator/runtime/src/apis.rs
```diff
@@ -483,4 +566,252 @@ impl_runtime_apis! {
             vec![]
         }
     }
+
+        impl fp_rpc::EthereumRuntimeRPCApi<Block> for Runtime {
+        fn chain_id() -> u64 {
+            <Runtime as pallet_evm::Config>::ChainId::get()
+        }
```
ahmadkaouk: Fix the indentation.

operator/node/src/client.rs
ahmadkaouk, line 1: Why adding this ? Is it really necessary ? I don't see it used anywhere.

gonzamontiel: Oh, I stated they were for later use in the [commit](https://github.com/Moonsong-Labs/datahaven/pull/13/commits/f8fdcd0a46a34dbdd4847e09d5e87441478cc032), forgot to add a comment, but I can delete them and re-add when needed also.

operator/runtime/src/apis.rs
gonzamontiel: Okay, I mean `proof_size_base_cost` and `weight_limit` are optional, but we still need to pass some `gas_limit` to `pallet_evm::Runner::create()`. I'll fix it.

gonzamontiel: Fixed.

PR #28, opened 2025-04-02
feat(contracts): 🚀 Implement deployment scripts for all contracts

ffarall:
In this PR:
1. Implement deployment script for all contracts involved. This includes
    1. EigenLayer core contracts.
    2. Snowbridge contracts.
    3. Our custom contracts to interact with both protocols.
4. Update README to document functionalities of `contracts` directory.

Future work:
1. Deployment of EigenLayer contracts should be conditional, as we wouldn't deploy them in testnet or mainnet, but we would do it in an internal stagenet.
5. Cleaning up unused smart contracts.


contracts/config/anvil.json
```diff
@@ -0,0 +1,52 @@
+{
```
timbrinded, line 1: Is there any write up on which of these values are important and what are boilerplate?

contracts/deployments/.gitkeep
```diff
@@ -0,0 +1 @@
+
```
timbrinded: what's this?

contracts/deployments/anvil.json
```diff
@@ -0,0 +1 @@
+{"network": "anvil","BeefyClient": "0x56fC17a65ccFEC6B7ad0aDe9BD9416CB365B9BE8","AgentExecutor": "0x2625760C4A8e8101801D3a48eE64B2bEA42f1E96","Gateway": "0xD6b040736e948621c5b6E0a494473c47a6113eA8","ServiceManager": "0xccf1769D8713099172642EB55DDFFC0c5A444FE9","VetoableSlasher": "0x3904b8f5b0F49cD206b7d5AABeE5D1F37eE15D8d","RewardsRegistry": "0x2Dd78Fd9B8F40659Af32eF98555B8b31bC97A351","Agent": "0xC79b5E0144570a8ec0F6634dbe95769c4521f42D","DelegationManager": "0xB06c856C8eaBd1d8321b687E188204C1018BC4E5","StrategyManager": "0xaB7B4c595d3cE8C85e16DA86630f2fc223B05057","AVSDirectory": "0xAD523115cd35a8d4E60B3C0953E0E0ac10418309","EigenPodManager": "0x045857BDEAE7C1c7252d611eB24eB55564198b4C","EigenPodBeacon": "0x5133BBdfCCa3Eb4F739D599ee4eC45cBCD0E16c5","RewardsCoordinator": "0x2b5A4e5493d4a54E717057B127cf0C000C876f9B","AllocationManager": "0x413b1AfCa96a3df5A686d8BFBF93d30688a7f7D9","PermissionController": "0x02df3a3F960393F5B349E40A599FEda91a7cc1A7"}
```
timbrinded: should this be gitignored?

contracts/foundry.toml
```diff
@@ -109,7 +109,8 @@
 [rpc_endpoints]
     mainnet = "${RPC_MAINNET}"
     holesky = "${RPC_HOLESKY}"
+    anvil = "http://localhost:8545"
 
-[etherscan]
-    mainnet = { key = "${ETHERSCAN_API_KEY}" }
-    holesky = { key = "${ETHERSCAN_API_KEY}" }
+# [etherscan]
```
timbrinded, line 114: intended?

contracts/src/libraries/MerkleUtils.sol
```diff
@@ -0,0 +1,180 @@
+// SPDX-License-Identifier: UNLICENSED
```
timbrinded, line 1: is this novel or copy paste from somewhere? Should this be unlicensed or MIT

contracts/deployments/.gitkeep
ffarall: It was initially there for git to keep this repo despite not having files. I could remove it now though.

contracts/deployments/anvil.json
ffarall: For Anvil deployment, yeah. The thing is that for other deployments like testnet/mainnet it could be very useful.

contracts/foundry.toml
ffarall: The `forge script` fails otherwise, saying that it didn't find the env var `ETHERSCAN_API_KEY`. I tried to tell it that I didn't care about that variable when running an anvil chain, but no luck. I accept suggestions, I want to deal with this some time in the future.

contracts/src/libraries/MerkleUtils.sol
ffarall: Actually, we should ask MBF what license we should be using. For now I think it is mostly irrelevant. 

cc @stiiifff

contracts/config/anvil.json
ffarall: Good point. Done.

PR #32, opened 2025-04-07
Add Snowbridge Inbound Queue V2 pallet

ahmadkaouk:
This PR adds the Snowbridge Inbound Queue V2 pallet to the runtime. It also defines bridge primitives, including the message structure for EigenLayer, and introduces the EigenLayerMessageProcessor to handle messages coming from EigenLayer.

### DIFF
--- a/operator/pallets/ethereum-client/Cargo.toml
+++ b/operator/pallets/ethereum-client/Cargo.toml
@@ -86,3 +86,8 @@ try-runtime = [
 	"pallet-timestamp?/try-runtime",
 	"sp-runtime/try-runtime",
 ]
+
+# THIS IS JUST TO AVOID TESTS RUNNING IN THE CI.
+# THIS IS A TEMPORARY PALLET HERE, AND THEN WE'LL IMPORT IT FROM REMOTE.
+[lib]
+test = false


operator/primitives/bridge/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "datahaven-bridge-primitives"
```
ffarall: How about renaming it to `dhp-bridge`? To follow the naming convention of `sp` for Substrate primitives and `fp` for Frontier primitives

operator/primitives/bridge/src/lib.rs
```diff
@@ -0,0 +1,84 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::pallet_prelude::*;
+use frame_system::RawOrigin;
+use parity_scale_codec::DecodeAll;
+use snowbridge_inbound_queue_primitives::v2::{Message as SnowbridgeMessage, MessageProcessor};
+use sp_std::vec::Vec;
+
+pub const EL_MESSAGE_ID: [u8; 4] = [112, 21, 0, 56];
```
ffarall, line 9: Is this arbitrary?

ahmadkaouk: I guess it is (I copied it from Tanssi). We need to make sure that this is unique to each message type.

PR #39, opened 2025-04-11
feat: ✨ Relay Validators operator set through Snowbridge

ffarall:
In this PR:
1. Implement application-specific functionalities in the `DataHavenServiceManager` contract:
    1. Registering of 3 operator sets: Validators, BSPs and MSPs.
    2. Allowlisted sign up of operators.
    3. Integration with Snowbridge to send message of new validator set.
2. Basic testing of the above functionalities.
3. Tests now use less mocked contracts (especially from EigenLayer).
4. Refactor of `SignUpOperator` script, which now supports the three kinds of Operator sets.


contracts/script/deploy/DeployLocal.s.sol
```diff
@@ -778,4 +717,113 @@ contract Deploy is Script, DeployParams, Accounts {
...
+        Logging.logStep("DataHaven service registered in AllocationManager");
+
+        // Set the slasher in the ServiceManager
+        vm.broadcast(_avsOwnerPrivateKey);
+        serviceManager.setSlasher(vetoableSlasher);
+        Logging.logStep("Slasher set in ServiceManager");
+
+        // Set the RewardsRegistry in the ServiceManager
+        vm.broadcast(_avsOwnerPrivateKey);
+        serviceManager.setRewardsRegistry(0, rewardsRegistry);
```
TDemeco: Now that we have multiple operator sets, instead of hardcoding the validator set to have ID = 0 here we can use an intermediate variable, just in case we end up changing this ID in the future. Something like:

```suggestion
        serviceManager.setRewardsRegistry(validatorSetOperatorSetId, rewardsRegistry);
```

contracts/script/transact/SignUpOperatorBase.s.sol
```diff
@@ -0,0 +1,164 @@
...
+            linkedToken.approve(address(strategyManager), balanceToStake);
+            strategyManager.depositIntoStrategy(deployedStrategies[i], linkedToken, balanceToStake);
+            vm.stopBroadcast();
+
+            Logging.logStep(
+                string.concat(
+                    "Staked ", vm.toString(balanceToStake), " tokens for strategy ", vm.toString(i)
+                )
+            );
+        }
```
TDemeco, line 102: Maybe we could have more flexibility here instead of deploying capital to all available strategies. Can the run function of the script receive arguments? If so, it could receive the list of indexes of the deployed strategies array to deploy capital into. If not, maybe we could use one of the `vm` functions, inheriting a config contract (similar to our inherit of the `Accounts` contract) or a constructor for that. Same with the amount of capital for each strategy

contracts/src/DataHavenServiceManager.sol
```diff
@@ -1,22 +1,55 @@
...
- * @title DataHaven ServiceManager contract.
- * TODO: For now, it uses the ServiceManagerBase contract as is.
- * TODO: We should add the DataHaven specific logic here.
+ * @title DataHaven ServiceManager contract
+ * @notice Manages validators, backup storage providers (BSPs), and main storage providers (MSPs)
+ * in the DataHaven network
  */
-contract DataHavenServiceManager is ServiceManagerBase {
-    uint256 public number;
+contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager {
```
TDemeco, line 31: Nice, it's finally taking shape! 💯

contracts/src/DataHavenServiceManager.sol
```diff
@@ -1,22 +1,55 @@
...
+    mapping(address => bool) public validatorsAllowlist;
+    /// @inheritdoc IDataHavenServiceManager
+    mapping(address => bool) public bspsAllowlist;
+    /// @inheritdoc IDataHavenServiceManager
+    mapping(address => bool) public mspsAllowlist;
+
+    IGatewayV2 private _snowbridgeGateway;
+
+    /// @inheritdoc IDataHavenServiceManager
+    mapping(address => bytes32) public validatorAddressToSolochainAddress;
```
TDemeco: Hmmm the naming here might be confusing since both are the validator's addresses, maybe something like:
```suggestion
    mapping(address => bytes32) public validatorEthAddressToSolochainAddress;
```
Although I don't like that either... Feel free to leave it as is, your call

contracts/src/DataHavenServiceManager.sol
```diff
@@ -25,23 +58,308 @@ contract DataHavenServiceManager is ServiceManagerBase {
...
+        OperatorSet memory operatorSet = OperatorSet({avs: address(this), id: VALIDATORS_SET_ID});
+        require(
+            _allocationManager.isMemberOfOperatorSet(msg.sender, operatorSet),
+            CallerIsNotValidator()
+        );
+        _;
+    }
+
+    /// @inheritdoc IDataHavenServiceManager
     function initialize(
```
TDemeco:
```suggestion
    function initialise(
```

contracts/src/DataHavenServiceManager.sol
```diff
@@ -25,23 +58,308 @@ contract DataHavenServiceManager is ServiceManagerBase {
...
+        DataHavenSnowbridgeMessages.NewValidatorSet memory newValidatorSetMessage =
+        DataHavenSnowbridgeMessages.NewValidatorSet({
+            nonce: 0,
+            topic: bytes32(0),
+            payload: newValidatorSetPayload
+        });
+
+        // Return the encoded message
+        return DataHavenSnowbridgeMessages.scaleEncodeNewValidatorSetMessage(newValidatorSetMessage);
+    }
```
TDemeco: Does it make sense for this function to be public or could it be internal? I don't mind it being public since it's gonna be useful for us when testing (and maybe if we at some point have to send manually a validator set update), but ideally if everything works ok I think it could be internal

contracts/src/DataHavenServiceManager.sol
```diff
@@ -25,23 +58,308 @@ contract DataHavenServiceManager is ServiceManagerBase {
...
+            if (!bspsAllowlist[operator]) {
+                revert OperatorNotInAllowlist();
+            }
+        }
+
+        if (operatorSetIds[0] == MSPS_SET_ID) {
+            if (!mspsAllowlist[operator]) {
+                revert OperatorNotInAllowlist();
+            }
+        }
```
TDemeco: This works and is clean but makes it a bit harder to add more operator sets in the future (since we would have to handle them in two places in the code). Something like this would make it a bit easier but I like both ways:
```suggestion
        if (operatorSetIds[0] == VALIDATORS_SET_ID) {
            if (!validatorsAllowlist[operator]) {
                revert OperatorNotInAllowlist();
            }

            // In the case of the Validators operator set, expect the data to have the Solochain address of the operator.
            // TODO: We should have some sort of validation of this address that validators are setting for themselves.
            validatorAddressToSolochainAddress[operator] = bytes32(data);
        } else if (operatorSetIds[0] == BSPS_SET_ID) {
            if (!bspsAllowlist[operator]) {
                revert OperatorNotInAllowlist();
            }
        } else if (operatorSetIds[0] == MSPS_SET_ID) {
            if (!mspsAllowlist[operator]) {
                revert OperatorNotInAllowlist();
            }
        } else {
          revert InvalidOperatorSetId();
        }
```

contracts/src/interfaces/IDataHavenServiceManager.sol
```diff
@@ -0,0 +1,282 @@
...
+    function bspsAllowlist(
+        address bsp
+    ) external view returns (bool);
+
+    /// @notice Checks if an MSP address is in the allowlist
+    /// @param msp Address to check
+    /// @return True if the MSP is in the allowlist, false otherwise
+    function mspsAllowlist(
+        address msp
+    ) external view returns (bool);
```
TDemeco, line 102: Super nitpick, these functions could use clearer names e.g. `isValidatorInAllowlist` or something like that

contracts/src/interfaces/IDataHavenServiceManager.sol
```diff
@@ -0,0 +1,282 @@
...
+    function snowbridgeGateway() external view returns (address);
+
+    /**
+     * @notice Converts a validator address to the corresponding Solochain address
+     * @param validatorAddress The address of the validator to convert
+     * @return The corresponding Solochain address
+     */
+    function validatorAddressToSolochainAddress(
+        address validatorAddress
+    ) external view returns (bytes32);
```
TDemeco: Same as the comment above^

contracts/src/interfaces/IDataHavenServiceManager.sol
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * @notice Initializes the DataHaven Service Manager
+     * @param initialOwner Address of the initial owner
+     * @param rewardsInitiator Address authorized to initiate rewards
+     * @param validatorsStrategies Array of strategies supported by validators
+     * @param bspsStrategies Array of strategies supported by BSPs
+     * @param mspsStrategies Array of strategies supported by MSPs
+     */
+    function initialize(
```
TDemeco:
```suggestion
    function initialise(
```

contracts/test/ServiceManagerBase.t.sol
```diff
@@ -46,17 +45,6 @@ contract ServiceManagerBaseTest is MockAVSDeployer {
...
-    function test_createOperatorSetsRevertsIfNoMetadataExists() public {
-        vm.prank(avsOwner);
-        vm.expectRevert(
-            abi.encodeWithSelector(IAllocationManagerErrors.NonexistentAVSMetadata.selector)
-        );
-
-        IAllocationManager.CreateSetParams[] memory emptyParams =
-            new IAllocationManager.CreateSetParams[](0);
-        ServiceManagerBase(address(serviceManager)).createOperatorSets(emptyParams);
-    }
```
TDemeco, line 58: Is this no longer true? Can a operator set be created even when the AVS metadata has not been set?

contracts/test/SnowbridgeIntegration.t.sol
```diff
@@ -5,19 +5,29 @@ pragma solidity ^0.8.13;
...
+import {OperatorSet} from "eigenlayer-contracts/src/contracts/libraries/OperatorSetLib.sol";
 
 import {MerkleUtils} from "../src/libraries/MerkleUtils.sol";
 import {
     IRewardsRegistryEvents, IRewardsRegistryErrors
 } from "../src/interfaces/IRewardsRegistry.sol";
-import {MockSnowbridgeAndAVSDeployer} from "./utils/MockSnowbridgeAndAVSDeployer.sol";
-
+import {SnowbridgeAndAVSDeployer} from "./utils/SnowbridgeAndAVSDeployer.sol";
+import {AVSDeployer} from "./utils/AVSDeployer.sol";
```
TDemeco: Do we need both?

contracts/src/DataHavenServiceManager.sol
ffarall: I couldn't use it in testing if it was internal 😬

contracts/src/interfaces/IDataHavenServiceManager.sol
ffarall: Fair point, they have that name because in that way they're automatically implemented by having the mappings be `public`. Essentially they have the name of the mapping.

contracts/src/interfaces/IDataHavenServiceManager.sol
ffarall: Same as above

contracts/test/SnowbridgeIntegration.t.sol
ffarall: Nope, good point.

contracts/test/ServiceManagerBase.t.sol
ffarall: The thing is that this is now handled in the initialise method of the DataHavenServiceManager.

contracts/script/transact/SignUpOperatorBase.s.sol
ffarall: Unfortunately it's not that easy. `forge script` runs by default the `run()` function without any parameter. You can make it run specific functions but it's a bit trickier from what I saw. In a future PR we could make this more flexible and configurable if needed, for sure.

PR #40, opened 2025-04-14
feat: add storage hub runtime

undercover-cactus:
In this PR we add StorageHub pallets to the runtime. It will only be added if the `storage-hub` feature is activated.


operator/Cargo.toml
```diff
@@ -147,3 +148,36 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
 fc-rpc = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2409", default-features = false }
 fc-rpc-core = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2409", default-features = false }
 fc-storage = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2409", default-features = false }
+
+# StorageHub
+storage-hub-runtime = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+pallet-aura = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
stiiifff: Is this required for SH ?

operator/Cargo.toml
```diff
@@ -147,3 +148,36 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
...
+pallet-message-queue = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
stiiifff: Is this related to XCM ?

operator/Cargo.toml
```diff
@@ -147,3 +148,36 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
...
+pallet-collator-selection = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
stiiifff: Is this required ? Not sure we want this for DH ..

operator/Cargo.toml
```diff
@@ -147,3 +148,36 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
...
+shp-treasury-funding = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-forest-verifier = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-file-key-verifier = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-data-price-updater = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+
+pallet-parameters = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
+pallet-nfts = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
+parachains-common = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
+parachain-info = { package = "staging-parachain-info", git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
stiiifff: DH is not a parachain .. not sure we want to import this.

operator/Cargo.toml
undercover-cactus: Good catch I have removed it.

operator/Cargo.toml
undercover-cactus: We actually don't use it so I have removed it.

operator/Cargo.toml
undercover-cactus: Not required. I have removed it.

operator/Cargo.toml
undercover-cactus: We don't need it so I have removed it.

operator/Cargo.toml
```diff
@@ -147,3 +148,32 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
...
+shp-file-metadata = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-traits = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-treasury-funding = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-forest-verifier = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-file-key-verifier = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+shp-data-price-updater = { git = "https://github.com/Moonsong-Labs/storage-hub.git", branch = "all-in-one", default-features = false }
+
+pallet-parameters = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
+pallet-nfts = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
+parachains-common = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
stiiifff: How about `parachains-common` ? Is it used ?

operator/runtime/src/configs/mod.rs
```diff
@@ -622,3 +671,652 @@ impl snowbridge_pallet_ethereum_client::Config for Runtime {
     type FreeHeadersInterval = ();
     type WeightInfo = ();
 }
+
+// Storage Hub
```
gonzamontiel: I'm thinking whether is clearer to maintain a separate mod file for storage hub configurations, given that they are only added behind a feature, and it makes the file a bit more complex to navigate.

operator/Cargo.toml
undercover-cactus: Removed too.

operator/runtime/src/configs/mod.rs
undercover-cactus: Done. It has now it own `storagehub` config folder.

operator/Cargo.toml
```diff
@@ -212,3 +213,31 @@ fc-mapping-sync = { git = "https://github.com/polkadot-evm/frontier", branch = "
 fc-rpc = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2412", default-features = false }
 fc-rpc-core = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2412", default-features = false }
 fc-storage = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2412", default-features = false }
+
+# StorageHub
+storage-hub-runtime = { git = "https://github.com/Moonsong-Labs/storage-hub.git", rev = "e32b6a8fbbdf241d19d211c3325be7329869b85d", default-features = false }
```
gonzamontiel, line 220: Is there a good reason we are using a rev instead of a branch ?

operator/runtime/stagenet/src/lib.rs
```diff
@@ -348,6 +387,159 @@ mod runtime {
     // ╚═══════════════════ DataHaven-specific Pallets ══════════════════╝
 }
 
+// Create the runtime by composing the FRAME pallets that were previously configured.
+#[frame_support::runtime]
+#[cfg(feature = "storage-hub")]
+mod runtime {
```
gonzamontiel, line 426: I assume we mean this to be temporary, but still, we would be generating 6 different versions of the runtime, stagenet, mainet, testnet, and each with or without `storage-hub` enabled. Looks hard to maintain. I see two options
1) We generate a common runtime mod with the types that sh needs, but only add the pallets behind the feature flag.
2) We add the pallets but we filter them out. In this case we would be compiling them, though.

operator/Cargo.toml
undercover-cactus: Yes because the main branch will change and that can break compatibility. We don't have tags so rev allow to fix the code.

gonzamontiel: I didn't mean the main branch. I meant a particular branch, for example called "stable2412", like we do with the rest of the dependencies in our repo, and also like Moonbeam does with all it's dependencies. A tag would also be better in my opinion that a rev. Not because I don't like revs, but because we need to be coherent with the rest of the repo and it's also more descriptive.

operator/runtime/testnet/Cargo.toml
```diff
@@ -101,6 +101,32 @@ xcm = { workspace = true }
...
+shp-traits = { workspace = true }
+shp-treasury-funding = { workspace = true }
+shp-forest-verifier = { workspace = true }
+shp-file-key-verifier = { workspace = true }
+shp-data-price-updater = { workspace = true }
+sp-trie = { workspace = true }
+## Misc
+smallvec = { workspace = true }
+num-bigint = { workspace = true }
+
```
ffarall: This shouldn't be here if it is not included in the `testnet` runtime.

operator/runtime/stagenet/Cargo.toml
```diff
@@ -101,6 +101,32 @@ xcm = { workspace = true }
...
+shp-file-metadata = { workspace = true }
+shp-traits = { workspace = true }
+shp-treasury-funding = { workspace = true }
+shp-forest-verifier = { workspace = true }
+shp-file-key-verifier = { workspace = true }
+shp-data-price-updater = { workspace = true }
+sp-trie = { workspace = true }
+## Misc
+smallvec = { workspace = true }
+num-bigint = { workspace = true }
```
ffarall: It doesn't make sense to me to have a separate section for these deps.

operator/runtime/stagenet/Cargo.toml
```diff
@@ -101,6 +101,32 @@ xcm = { workspace = true }
 xcm-builder = { workspace = true }
 xcm-executor = { workspace = true }
 
+# StorageHub
+## StorageHub and Polkadot deps
```
ffarall: I support separating the dependencies in sections. But this file doesn't do that. Personally, I'd suggest not to separate them in this PR, and then having another PR that separates in sections the dependencies in this Cargo.toml, just like in the other parts of this project.

operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -44,6 +49,308 @@ pub mod dynamic_params {
     }
 }
 
+#[cfg(feature = "storage-hub")]
+#[dynamic_params(RuntimeParameters, pallet_parameters::Parameters::<Runtime>)]
```
ffarall, line 58: I'd check if it is possible to not have the parameters duplicated here. In theory, having extra parameters that you don't use then shouldn't be a problem. I understand that you needed to do this because you made a separate file for the runtime with all things SH, and put the compilation of that file behind a feature flag.

In my opinion, the simplest way to do this would be to implement a `CustomCallFilter` such as this:
```
/// Custom call filter to disable specific pallet extrinsics
pub struct CustomCallFilter;
impl frame_support::traits::Contains<RuntimeCall> for CustomCallFilter {
    fn contains(call: &RuntimeCall) -> bool {
        match call {
            // Block all DataHavenNativeTransfer pallet calls
            RuntimeCall::YourPallet(_) => false,
            // Allow all other calls
            _ => true,
        }
    }
}
```

This way, we can still have just ONE runtime definition, but still have the pallets disabled because you cannot interact with them.

ffarall: Then in the `frame_system::Config` trait implementation where you configure the system pallet, you add this:
```
    /// Custom call filter to disable specific pallet extrinsics
    type BaseCallFilter = crate::CustomCallFilter;
```

And that you put behind the feature flag, so that if `storage-hub` feature is not used, the `CustomFilter` is used.

operator/runtime/common/src/lib.rs
```diff
@@ -43,6 +43,9 @@ pub type Nonce = u32;
 /// A hash of some data used by the chain.
 pub type Hash = sp_core::H256;
 
+/// The hashing algorithm used.
+pub type Hashing = BlakeTwo256;
+
```
ffarall: What is this for? For SH? Probably we could use a more descriptive name if it's just used for SH.

operator/runtime/stagenet/src/configs/runtime_params.rs
gonzamontiel: If we don't care about _compiling_ everything, agreed this is the simplest approach 💯

operator/runtime/testnet/Cargo.toml
undercover-cactus: I forgot to remove them. It was left over from merging with main.

I have removed them.

operator/runtime/stagenet/Cargo.toml
undercover-cactus: It was to not forget to make them optional. It helps with saving build time. I have marked them as optional (and the rest of the storagehub deps).

operator/runtime/stagenet/Cargo.toml
undercover-cactus: This has been removed.

operator/runtime/stagenet/src/configs/runtime_params.rs
undercover-cactus: I tried this solution but it wouldn't compile. I didn't spend time debugging it because the way it is being done currently works and it is temporary.

Happy to do it in another PR once I am done with the higher priority task

operator/runtime/common/src/lib.rs
undercover-cactus: Indeed we need a Hashing type defined for Storage Hub. Looking at the project we actually use different type of hashing functions for different reason (e.g Ethereum use keccak) so it doesn't make sense to have it in common.

I have move it definition to the `storage-hub` config folder.

operator/runtime/stagenet/src/configs/runtime_params.rs
ffarall: Ok, given that it's temporary, let's go ahead with that.

PR #41, opened 2025-04-15
feat: 🧑‍💻 Turn scripts into an interactive CLI

ffarall:
WARNING: This PR changes the kurtosis package to use the one from upstream, not our fork, as it was not working at the moment. This should be changed when fixed.

In this PR:
1. Turn `launch-kurtosis` script into a CLI, which parses parameters and can interactively run multiple steps.
2. Separate steps of such CLI into their own scripts.
    1. New script created `launch-kurtosis`, which detects if an enclave is running and prompts to relaunch it if it is.
    2. New script created `deploy-contracts` to deploy all contracts. It can optionally verify them as well.
3. Each step can be interactively opted in/out, or choose whether to run it via CLI params.
4. The CLI offers a help command as well.
5. Cleanup logs of CLI. Logs of internal commands ran can be printed with LOG_LEVEL=debug. In case of error, they are always printed out.

### DIFF
--- a/test/scripts/launch-kurtosis.ts
+++ b/test/scripts/launch-kurtosis.ts
@@ -51,6 +51,7 @@ export const launchKurtosis = async (options: { launchKurtosis?: boolean } = {})
   logger.info("🧹 Cleaning up Docker and Kurtosis environments...");
   logger.debug(await $`kurtosis enclave stop datahaven-ethereum`.nothrow().text());
   logger.debug(await $`kurtosis clean`.text());
+  logger.debug(await $`kurtosis engine stop`.text());
   logger.debug(await $`docker system prune -f`.nothrow().text());
 
   // Pull necessary Docker images


test/scripts/deploy-contracts.ts
```diff
@@ -0,0 +1,145 @@
...
+    throw Error("❌ Contracts have failed to build properly.");
+  }
+  logger.debug(buildStdout.toString());
+
+  // Get forge path
+  const { stdout: forgePath } = await $`which forge`.quiet();
+  const forgeExecutable = forgePath.toString().trim();
+
+  // Prepare deployment command
+  let deployCommand = `${forgeExecutable} script script/deploy/DeployLocal.s.sol --rpc-url ${rpcUrl} --color never -vv --no-rpc-rate-limit --non-interactive --broadcast`;
```
stiiifff, line 71: I got an error when running `start:e2e:verified` or `start:e2e:minimal`.
Looks like `DeployLocal.s.sol` is not the right contract name.

image

ffarall: It should be though. Have you tried doing forge clean before? I've seen forge sometimes do weird things with the cached builds when the last build you've done had different contract names.

timbrinded: yes i got this too steve, if you cd to `/contracts` and `forge clean` like facu mentioned that should fix it

stiiifff: Yep, that worked !

test/scripts/e2e-cli.ts
```diff
@@ -0,0 +1,190 @@
...
+  ${chalk.gray("# Start with verification and automatic redeploy")}
+  bun run start-kurtosis --verified --redeploy
+
+  ${chalk.gray("# Start without deploying contracts")}
+  bun run start-kurtosis --no-deploy-contracts
+`);
+}
+
+// Parse and handle boolean flags with negations
+function parseFlag(args: string[], flagName: string): boolean | undefined {
```
timbrinded, line 181: consider using https://nodejs.org/api/util.html#utilparseargsconfig

timbrinded: or https://www.npmjs.com/package/commander

test/utils/input.ts
```diff
@@ -0,0 +1,51 @@
+import chalk from "chalk";
```
timbrinded, line 1: this is impressive but probably safer just to use https://www.npmjs.com/package/@inquirer/prompts

test/scripts/e2e-cli.ts
ffarall: Good suggestion! I can do the refactor in a later PR 👌🏼

test/utils/input.ts
ffarall: Also good suggestion! I'll do it in the refactor as well

PR #47, opened 2025-04-21
feat(contracts): ✨ add set up validator script and execute it when starting integration tests

TDemeco:
This PR adds the `setup-validators` Typescript script that, given an already started up network, sets up a new validator set and sends it through Snowbridge's Gateway to the solochain. To accomplish that purpose, this PR:
- Modifies the `DeployLocal` script to save in the `anvil.json` file not only the deployed strategies' addresses but also the owner of each strategy's underlying token. This owner is used as the source of funds to transfer tokens to other validators so they can register under that strategy.
- Adds an `OPERATOR_SOLOCHAIN_ADDRESS` to the `Accounts` utility script contract. This address is the one used as the Solochain address when registering a new Operator.
- Updates the `SignUpOperator` (which I believe is now deprecated since we have multiple Operator Sets) and `SignUpOperatorBase` scripts to adapt to both aforementioned changes.
- Updates the `ELScriptStorage` script to save the new extra information of each deployed strategy (the creator of the underlying token) in storage.
- Adds a `validator-set.json` file which contains the validators that should be registered in the AVS and sent to the Solochain network through the Snowbridge Gateway when starting any integration test. This is currently hardcoded but could be generated in any other way, giving us flexibility for testing.
- Adds both a Markdown file and a Excalidraw diagram showcasing both how the setup of integration tests work and possible integration tests that will be added in a future PR. This list is not exahustive as there are many more scenarios we will want to test using integration tests.
- Updates the `e2e-cli.ts` script to execute the validator setup when bootstrapping the network used for integration testing.


test/scripts/setup-validators.ts
```diff
@@ -0,0 +1,444 @@
...
+  const forgeExecutable = forgePath.toString().trim();
+
+  // Get cast path for transactions
+  const { stdout: castPath } = await $`which cast`.quiet();
+  const castExecutable = castPath.toString().trim();
+
+  // 1. Register validators in EigenLayer using SignUpValidator.s.sol
+  if (shouldExecuteSignup) {
+    // Get the deployment information to find the strategies
+    const deploymentPath = path.resolve("../contracts/deployments/anvil.json");
```
ffarall: This can actually change. `anvil.json` is just the default file name.

test/scripts/setup-validators.ts
```diff
@@ -0,0 +1,444 @@
...
+      const creatorEthBalance = ethBalanceOutput.toString().trim();
+      logger.info(`Token creator has ${creatorErc20Balance} tokens and ${creatorEthBalance} ETH`);
+      //const getBalanceCmd = `${castExecutable} call ${underlyingTokenAddress} "balanceOf(address)(uint256)" ${tokenCreator} --rpc-url ${rpcUrl}`;
+      //const { stdout: balanceOutput } = await $`sh -c ${getBalanceCmd}`.quiet();
+      //const creatorBalance = balanceOutput.toString().trim().split(" ")[0];
+      //logger.info(`Token creator has ${creatorBalance} tokens`);
+
+      // Transfer 5% of the creator's tokens to each validator + 1% of the creator's ETH. ETH is transferred only if the receiving validator does not have any
+      const erc20TransferAmount = BigInt(creatorErc20Balance) / BigInt(20); // 5% of the balance
+      const ethTransferAmount = BigInt(creatorEthBalance) / BigInt(100); // 1% of the balance
```
ffarall: Not sure this is the behaviour the script should have. Assuming the token creator is one of the validators only makes sense when we are testing, and both are pre-funded accounts. In a more real world scenario, the tokens would be pre-existing tokens, for which the validators should already have balance. I think this script should be closer to that. 

Let me know what you think and whether we should discuss.

test/scripts/setup-validators.ts
```diff
@@ -0,0 +1,444 @@
...
+      const getEthBalanceCmd = `${castExecutable} balance ${tokenCreator} --rpc-url ${rpcUrl}`;
+      const { stdout: erc20BalanceOutput } = await $`sh -c ${getErc20BalanceCmd}`.quiet();
+      const { stdout: ethBalanceOutput } = await $`sh -c ${getEthBalanceCmd}`.quiet();
+      const creatorErc20Balance = erc20BalanceOutput.toString().trim().split(" ")[0];
+      const creatorEthBalance = ethBalanceOutput.toString().trim();
+      logger.info(`Token creator has ${creatorErc20Balance} tokens and ${creatorEthBalance} ETH`);
+      //const getBalanceCmd = `${castExecutable} call ${underlyingTokenAddress} "balanceOf(address)(uint256)" ${tokenCreator} --rpc-url ${rpcUrl}`;
+      //const { stdout: balanceOutput } = await $`sh -c ${getBalanceCmd}`.quiet();
+      //const creatorBalance = balanceOutput.toString().trim().split(" ")[0];
+      //logger.info(`Token creator has ${creatorBalance} tokens`);
```
ffarall: What's up with these commented lines?

test/scripts/setup-validators.ts
ffarall: Perhaps separating this "funding" step into a different script? That would be optionally run if we're deploying to a local network

ffarall: On the other hand, in this script used to deploy to Kurtosis, it is always a new, local network... Need to think about this.

test/scripts/setup-validators.ts
```diff
@@ -0,0 +1,444 @@
...
+
+    // Iterate through all validators to register them
+    for (let i = 0; i < validators.length; i++) {
+      const validator = validators[i];
+      logger.info(`Setting up validator ${i} (${validator.publicKey})`);
+
+      // Setting up the environment variables directly
+      const env = {
+        ...process.env,
+        NETWORK: "anvil",
```
ffarall: Should we hardcode this?

test/scripts/setup-validators.ts
```diff
@@ -0,0 +1,444 @@
...
+      }
+
+      logger.success(`Successfully registered validator ${validator.publicKey}`);
+    }
+  } else {
+    logger.info("Skipping validator registration in EigenLayer");
+  }
+
+  // 2. Send the validator set to the DataHaven chain through Snowbridge
+  if (shouldSendToSubstrate) {
```
ffarall: Worth considering if this should be it's own separate script. Especially once we involve the relayer and substrate, if this script proceeds to checking the message was sent.

ffarall: In fact, making it it's own separate script would allow us to pass different flags in the CLI to execute one or the other step.

test/resources/datahaven-integration-test-flow.md
```diff
@@ -0,0 +1,208 @@
...
+  - VetoableSlasher: Manages slashing with veto capabilities
+
+### Initial Configuration
+
+- Initialize contracts with test accounts
+- Configure DataHavenServiceManager with initial operator sets
+- Configure Gateway with appropriate parameters
+- Configure RewardsRegistry with initial values
+
+### Key Commands
```
ffarall, line 72: We could mention that you can run the CLI avoiding the previous steps of deployment.

test/scripts/setup-validators.ts
ffarall, line 1: This script is too verbose in my opinion. Some logs can be turned into `debug`, and the output of the transactions can also be run with `logger.debug($'command'.text())`

test/scripts/setup-validators.ts
gonzamontiel: Instead, we can rely on a `.env.template` with the default values

test/scripts/setup-validators.ts
TDemeco: That's true. I changed the script to select the `anvil` network by default, but it now allows for both passing a custom path for the deployments JSON or a network name (which would use the same path as `anvil` but with a different network name)

test/scripts/setup-validators.ts
TDemeco: I separated this into its own script that you can select to run when running the whole E2E test setup, and is run by default (since we are deploying a new network)

test/scripts/setup-validators.ts
TDemeco: Sorry, removed! I was testing something

test/scripts/setup-validators.ts
TDemeco: NETWORK is now selectable when running both the E2E setup and the standalone script. The rest is specific `.env` changes that we do to pass information to the Solidity script (which validator to register and what's his solochain address)

test/scripts/setup-validators.ts
TDemeco: Yep, I've extracted this into its own script

test/resources/datahaven-integration-test-flow.md
TDemeco: Updated the docs to reflect the new separation of scripts and added a disclaimer about being able to run the scripts independently

test/scripts/setup-validators.ts
TDemeco: I changed a few of the `info` logs to `debug` to avoid the script being too verbose, but let me know if you still find it overly verbose

PR #50, opened 2025-04-22
feat(operator): :sparkles: Add pallet params and use it for `EthereumfGatewayAddress`

ffarall:
This PR adds the [pallet parameters](https://docs.rs/pallet-parameters/0.11.0/pallet_parameters/index.html) to the DataHaven runtime.

This allows config parameters for pallets to be dynamically modified, needed, for instance, to set `EthereumfGatewayAddress` after the deployment of that smart contract on Ethereum. StorageHub also uses this pallet extensively.

Additionally, refactors the pallet configuration files to have them more visually organised.


operator/runtime/src/configs/mod.rs
```diff
@@ -83,6 +79,16 @@ use sp_std::{
 };
 use sp_version::RuntimeVersion;
 
+use super::{
```
gonzamontiel, line 82: Do you use a formatter for auto-organizing the imports? If you do, it would be nice that we all use the same convention :)

operator/runtime/src/configs/mod.rs
```diff
@@ -98,6 +104,20 @@ pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(WEIGHT_REF_TIME_PER_
 
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
 
+//╔═══════════════════════════════════════════════════════════════════════════════════════════════════════════════╗
```
gonzamontiel, line 107: fancy ✨

operator/runtime/src/configs/runtime_params.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub mod runtime_config {
+
+        use super::*;
+
+        #[codec(index = 0)]
+        #[allow(non_upper_case_globals)]
+        /// Set the initial address of the Snowbridge Gateway contract on Ethereum.
+        /// The fact that this is a parameter means that we can set it initially to the zero address,
+        /// and then change it later via governance, to the actual address of the deployed contract.
+        pub static EthereumGatewayAddress: H160 = H160::repeat_byte(0x0);
```
gonzamontiel, line 20: I guess the pallet inbound message has some sort of check agains a zero address? Before, I see we were setting it to `0x42`, do you know why ? Anyway, looks safe.

operator/runtime/src/configs/mod.rs
ffarall: Not really, I usually follow this structure of imports:
```
std imports

other packages imports

crate/super imports
```

I do it manually though, I don't have a formatter, but I've been thinking of having some common Cursor rules so that the Agent tries to "enforce" this order. Never get the time to do it though, and it would only be useful if you're on Cursor.

operator/runtime/src/configs/runtime_params.rs
ffarall: Not sure it does, but this is because we don't know in compile time the address of the deployed Gateway contract, so I thought we might as well set it to the 0 address, and then we can change it as it is a parameter.

PR #52, opened 2025-04-23
test: 🧪 Initial Relay Work

timbrinded:
Small changes to help progress us closer to relayer support


test/utils/docker.ts
```diff
@@ -123,3 +123,5 @@ export const getPublicPort = async (
   invariant(portMappings, `❌ port mapping not found for ${containerName}:${internalPort}`);
   return portMappings.PublicPort;
 };
+
+export const startRelayer = async () => {};
```
ffarall: Does it make sense for this to be here?

operator/node/src/chain_spec.rs
```diff
@@ -41,7 +41,7 @@ pub fn local_chain_spec() -> Result<ChainSpec, String> {
         None,
     )
     .with_name("Local Testnet")
-    .with_id("local_testnet")
+    .with_id("local")
```
ffarall, line 44: Why?

timbrinded: To that it fulfils the match arm here:
https://github.com/Moonsong-Labs/datahaven/blob/165736aeeecdb5427cbf2641868f38009e9db4f0/operator/node/src/command.rs#L43

test/utils/docker.ts
timbrinded: was placeholder for later but yeah lets kill it

PR #58, opened 2025-04-30
test: 🧙 Generate Type Bindings for Contracts

timbrinded:
## Summary
This PR adds statically typed bindings for contracts. This allows you to write E2E tests with full completions in TS.

## Additions

- `ts-build.yml` New CI, this will make sure that if there's changes made to the contracts that the contract-bindings are up to date.
- `package.json` script changes
  - `start:e2e:ci` - Designed to be run with all options specified since CIs are famously bad with iteractive CLI prompts
  - `test:e2e` - added timeout
  - `generate:wagmi` - This generates the smart contract bindings for our tests
- New Function Helpers:
  - `generateRandomAccount()` Returns a viem account type object for a random account. Useful for tests where you want idempotency on a long lived network since the state is probabilistically fresh
  - `getContractInstance()` Returns a viem contract instance that allows you to read/write to the deployed contract. You should get full type inference here for the methods available and parameters required.

### Example

```ts
 it("avs() can be read from contract instance", async () => {
    const value = await instance.read.avs();
    expect(isAddress(value), "AVS getter should return an address").toBeTrue();
  });
```

### DIFF
--- a/test/cli/handlers/launch/launchedNetwork.ts
+++ b/test/cli/handlers/launch/launchedNetwork.ts
@@ -4,7 +4,7 @@ import { logger } from "utils";
 
 export class LaunchedNetwork {
   protected runId: string;
-  protected processes: Bun.Subprocess[];
+  protected processes: Bun.Subprocess<"inherit" | "pipe" | "ignore", number, number>[];
   protected fileDescriptors: number[];
   protected DHNodes: { id: string; port: number }[];
 
@@ -33,7 +33,7 @@ export class LaunchedNetwork {
     this.fileDescriptors.push(fd);
   }
 
-  addProcess(process: Bun.Subprocess) {
+  addProcess(process: Bun.Subprocess<"inherit" | "pipe" | "ignore", number, number>) {
     this.processes.push(process);
   }
 


test/utils/contracts.ts
```diff
@@ -0,0 +1,152 @@
...
+  type Client,
+  createClient,
+  getContract,
+  isAddress
+} from "viem";
+import { http, createConfig } from "wagmi";
+import { mainnet, sepolia } from "wagmi/chains";
+import { z } from "zod";
+import { logger } from "./logger";
+import { type ViemClientInterface, createChainConfig, createDefaultClient } from "./viem";
```
ffarall: A bunch of imports here are never used. Is this on purpose?

test/utils/contracts.ts
```diff
@@ -0,0 +1,152 @@
...
+  AVSDirectory: generated.avsDirectoryAbi,
+  EigenPodManager: generated.eigenPodManagerAbi,
+  EigenPodBeacon: generated.eigenPodAbi,
+  RewardsCoordinator: generated.rewardsCoordinatorAbi,
+  AllocationManager: generated.allocationManagerAbi,
+  PermissionController: generated.permissionControllerAbi,
+  ETHPOSDeposit: generated.iethposDepositAbi,
+  BaseStrategyImplementation: generated.strategyBaseTvlLimitsAbi,
+  network: generated.beefyClientAbi, // placeholder
+  DeployedStrategies: generated.beefyClientAbi // placeholder
```
ffarall: What is `network` supposed to be a placeholder for?

test/package.json
```diff
@@ -8,16 +8,18 @@
     "fmt": "biome check .",
     "fmt:fix": "biome check --write .",
     "build:docker:relayer": "bun -e \"import build  from './scripts/snowbridge-relayer.ts'; build()\"",
+    "generate:wagmi": "wagmi generate",
```
ffarall, line 11: Do you think it's worth adding running this step to the test docs?

timbrinded: done

test/utils/contracts.ts
timbrinded: removed

test/utils/contracts.ts
timbrinded: was being lazy, this got dragged in since it's part of the `anvil.json` file, but have just omitted the key now.

PR #59, opened 2025-04-30
test: ⚡️ CI Refactor

timbrinded:
Eventually our CI will be required to run two private blockchains locally plus associated relayers. 

This PR is to prepare for this fate by improving run times and refactoring our existing CIs so they are a bit easier to reason about.

### Refactors
- **_We now run ALL CIs on every PR!_** This is so that we decomplexify the logic around conditional builds and fetching built binaries from another source. This reduces the surface area of code we have to maintain at the cost of execution time
  - This penalty is ameliorated by a layered caching system. At best, it will be less than a minute to complete a build since everything will be cached. On GH runners this is about 6 minutes sadly.
  - We will no longer be at risk of important CIs being skipped erroneously which hide true failures.
  - Caching is a low-risk approach because at worst it has to build from scratch. A bad cache hit will never imply the wrong thing gets build since cargo is smart enough to just throw away any inappropriate build artefacts.
- `setup-rust` action created so we have a unified way of setting up runner and unifying our approach to caching
  - Use a unique caching key for different activities and it will fallback to shared cache if no matches
- we are using `mainnet` kurtosis config so that it works with relayer assumptions

### Additions
- We can specify the ethereum block time via a new cli arg `--slot-time <seconds>`
- We can specify arbitrary network_param args which get passed into the generated yaml
  - e.g. giving `bun cli --kurtosis-network-args="pet=cat food=fish" will add:

```yml
network_params:
  # existing params...
  pet: cat
  food: fish
```

- We now have the ability to programmatically modify the yaml
  - This means we are back down to a single `minimal.yml` kurtosis config so we dont have to maintain changes between them
  - Flow is: `add new cli arg` -> `add if() block which mutates yaml` -> `profit`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,5 +1,6 @@
+#! Main CI Specification for DataHaven Repository
+
 name: CI
-description: "Main CI Specification for DataHaven Repository"
 
 on:
   workflow_dispatch:
@@ -14,11 +15,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-
   # First Tier
   build-operator:
     uses: ./.github/workflows/task-build-operator.yml
-    secrets: inherit # For demonstrative purposes, we dont use any secrets yet.
+    secrets: inherit # For demonstrative purposes, we don't use any secrets yet.
   ts-build:
     uses: ./.github/workflows/task-ts-build.yml
   ts-lint:
--- a/test/cli/handlers/launch/index.ts
+++ b/test/cli/handlers/launch/index.ts
@@ -133,15 +133,7 @@ export const launch = async (options: LaunchOptions) => {
 export const launchPreActionHook = (
   thisCmd: Command<[], LaunchOptions & { [key: string]: any }>
 ) => {
-  const {
-    blockscout,
-    verified,
-    fundValidators,
-    setupValidators,
-    updateValidatorSet,
-    deployContracts,
-    slotTime
-  } = thisCmd.opts();
+  const { blockscout, verified, fundValidators, setupValidators, deployContracts } = thisCmd.opts();
   if (verified && !blockscout) {
     thisCmd.error("--verified requires --blockscout to be set");
   }
--- a/.github/workflows/task-rust-lint.yml
+++ b/.github/workflows/task-rust-lint.yml
@@ -54,6 +54,7 @@ jobs:
         run: SKIP_WASM_BUILD=1 cargo clippy --features try-runtime,runtime-benchmarks --locked --release
 
   check-cargo-sort:
+    name: "Check Cargo sort"
     runs-on: ubuntu-latest
 
     defaults:


test/configs/kurtosis/minimal.yaml
```diff
@@ -14,6 +7,24 @@ additional_services:
...
+  seconds_per_slot: 6
   num_validator_keys_per_node: 128
-  prefunded_accounts: '{"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": {"balance": "10ETH"}, "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc": {"balance": "10ETH"}, "0x70997970C51812dc3A010C7d01b50e0d17dc79C8": {"balance": "10ETH"},"0x976ea74026e726554db657fa54763abd0c3a0aa9": {"balance": "10ETH"}}'
\ No newline at end of file
+  prefunded_accounts: '{
+      "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": {"balance": "10ETH"}, 
+      "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc": {"balance": "10ETH"}, 
+      "0x70997970C51812dc3A010C7d01b50e0d17dc79C8": {"balance": "10ETH"},
+      "0x976ea74026e726554db657fa54763abd0c3a0aa9": {"balance": "10ETH"}
+    }'
```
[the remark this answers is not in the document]
timbrinded: This breaks kurtosis, they cant process block scalars it looks like

test/configs/kurtosis/minimal.yaml
```diff
@@ -14,6 +7,26 @@ additional_services:
...
+  # Preloaded with DataHavenTest contract
+  additional_preloaded_contracts: |
```
timbrinded, line 20: You might find this part interesting @ffarall 

We can pre-deploy a contract with arbitrary storage slots already set.

We can use this for contract setup if we want to override certain values of the AVS if need be

test/package.json
```diff
@@ -10,9 +10,9 @@
     "build:docker:relayer": "bun -e \"import build  from './scripts/snowbridge-relayer.ts'; build()\"",
     "generate:wagmi": "wagmi generate",
     "generate:snowbridge-cfgs": "bun -e \"import {generateSnowbridgeConfigs} from './scripts/gen-snowbridge-cfgs.ts'; await generateSnowbridgeConfigs()\"",
-    "start:e2e:verified": "bun cli --verified --blockscout --deploy-contracts",
-    "start:e2e:minimal": "bun cli",
-    "start:e2e:ci": "bun cli -d --setup-validators --update-validator-set --fund-validators",
+    "start:e2e:verified": "bun cli --verified --blockscout --deploy-contracts --setup-validators --update-validator-set --fund-validators --slot-time 1",
+    "start:e2e:minimal": "bun cli --slot-time 2",
```
ffarall: I'd just make 2 the default to be honest

test/cli/handlers/launch/index.ts
```diff
@@ -138,7 +139,8 @@ export const launchPreActionHook = (
     fundValidators,
     setupValidators,
     updateValidatorSet,
-    deployContracts
+    deployContracts,
+    slotTime
```
ffarall: Linter says `updateValidatorSet` and `slotTime` are not used. Any reason to keep them?

test/utils/shell.ts
ffarall, line 1: This is VERY cool.

test/configs/kurtosis/minimal.yaml
ffarall: If we had a way of autogenerating this based on deployed contracts, meaning run the deploy and have it autogenerate this, it would be a massive speedup on our CI.

ffarall: I wouldn't do it manually though.

.github/workflows/actions/setup-env/action.yml
ffarall, line 1: Why is this named "action"?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,38 @@
+name: CI
+description: "Main CI Specification for DataHaven Repository"
```
ffarall: Linter says: `Property description is not allowed`

.github/workflows/CI.yml
ffarall, line 1: The new CI organisation looks beautiful

.github/workflows/actions/setup-env/action.yml
timbrinded: Because github makes you do so

test/utils/shell.ts
timbrinded: still cant get it to print the in-flight details when contracts are deployed though!

PR #65, opened 2025-05-07
feat(operator): Add external validators Pallet

ahmadkaouk:
This PR replaces `pallet-validator-set` with a new `pallet-external-validators` pallet from Tanssi.

## Key Changes

- **New ExternalValidators Pallet**
  - Supports whitelisted validators (set by governance, not rewarded)
  - Manages external validators (can be enabled/disabled)
  - Implements era-based rotation (eras change after configurable sessions)

- **Bridge Integration**
  - Updated `EigenLayerMessageProcessor` in `operator/primitives/bridge/src/lib.rs`
  - Replaced the old `SetValidators` command with  
    ```rust
    ReceiveValidators { validators, external_index }
    ```


operator/runtime/stagenet/src/lib.rs
```diff
@@ -318,5 +314,8 @@ mod runtime {
     // Start with index 100
     #[runtime::pallet_index(100)]
     pub type OutboundCommitmentStore = pallet_outbound_commitment_store;
+
+    #[runtime::pallet_index(101)]
+    pub type ExternalValidators = pallet_external_validators;
```
stiiifff: This still needs to be before the Session pallet, otherwise it cannot get the initial validator set, and the node crashes on startup.

ahmadkaouk: fixed in f178c53277cf97c915e33ab82b6a00d83a5fe17b

operator/runtime/stagenet/src/lib.rs
```diff
@@ -227,32 +227,31 @@ mod runtime {
...
-    pub type Historical = pallet_session::historical;
-
-    // Validator set must be before Session.
+    // External Validators must be before Session.
     #[runtime::pallet_index(7)]
-    pub type ValidatorSet = pallet_validator_set;
+    pub type ExternalValidators = pallet_external_validators;
 
     #[runtime::pallet_index(8)]
+    pub type Historical = pallet_session::historical;
```
stiiifff: Maybe not a biggie but Historical pallet is at index 6 in other runtimes .. for consistency sake, I would do the same here.

PR #67, opened 2025-05-12
fix: 🩹 update to allow BEEFY relayer to run correctly

TDemeco:
This PR:
- Adds a check to make sure the BEEFY RPC endpoint is ready before spinning up the BEEFY relayer, otherwise it would just fail and crash.
- Adds the `--enable-offchain-indexing=true` flag to the Datahaven nodes run when starting up the E2E infra. This is needed because otherwise nodes can't be queried by the relayer to generate the required proofs since they would not store the MMR leafs/nodes/root, so the relayer would just crash.
- Updates the way we were generating the merkle root of the validator set.
  - The BEEFY pallet (and as such, the relayer) generate the validator set merkle tree by getting the validator list and treating it as an already ordered set of merkle leafs, hashing each pair in succession without caring what each leaf value is.
  - Meanwhile, the OpenZeppelin crypto library (and as such, the EigenLayer contracts) also gets the merkle leafs list but hashes each pair of leafs in value order. 
  - This means that, for example, if the list of leafs is: ["0x124", "0x123"]
    - BEEFY would do `hash("0x124123")` to get the value of the parent node.
    - OZ and EigenLayer would do `hash("0x123124")` to get the value of the parent node.
  - This created a mismatch between what the BEEFY relayer was expecting and what was actually calculated in our script. A way to obtain a merkle tree using the BEEFY way was added to solve this.
- Updates the authority set to not be a hardcoded array of keys, now the BEEFY keys are obtained by directly querying the runtime before deploying the BEEFY contracts.
- Renames a few files from `flamingo` to `datahaven`.

## Summary by CodeRabbit

- **New Features**
	- Added support for calculating and generating Merkle roots and proofs without sorting, enabling new use cases for validator set management.
	- Introduced dynamic fetching and configuration of Datahaven validator authorities during network launch.
	- Added readiness check for the BEEFY protocol before relayer startup, improving reliability of relayer operations.

- **Bug Fixes**
	- Fixed indentation issues in configuration files for improved readability and consistency.

- **Chores**
	- Updated validator lists and addresses in configuration files.
	- Enhanced e2e test scripts and added new commands for relayer and minimal test scenarios.
	- Added new dependency to test package.
	- Updated version strings in package metadata.

- **Refactor**
	- Improved logging and configuration handling during network and relayer launches.
	- Simplified import statements and removed unused code.

- **Style**
	- Reformatted configuration and TypeScript files for better readability.


test/cli/handlers/launch/relayer.ts
```diff
@@ -109,9 +187,8 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
 
     const ethWsPort = await getPortFromKurtosis("el-1-reth-lighthouse", "ws");
     const ethHttpPort = await getPortFromKurtosis("cl-1-lighthouse-reth", "http");
-    const substrateWsPort = 9944;
     logger.debug(
-      `Fetched ports: ETH WS=${ethWsPort}, ETH HTTP=${ethHttpPort}, Substrate WS=${substrateWsPort} (hardcoded)`
+      `Fetched ports: ETH WS=${ethWsPort}, ETH HTTP=${ethHttpPort}, Substrate WS=${substrateWsPort} (from DataHaven node)`
     );
 
     if (type === "beacon") {
```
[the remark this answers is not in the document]
TDemeco: No need to handle, if Kurtosis is not running at this point it's a critical error and we can throw.

test/cli/handlers/launch/relayer.ts
```diff
@@ -17,6 +20,47 @@
...
+  } catch (error) {
+    logger.error(
+      `❌ Failed to connect to DataHaven node on port ${port} for BEEFY check: ${error}`
+    );
+    if (client) {
+      await client.destroy(); // Ensure client is destroyed on outer catch too
+    }
+    return false;
+  }
+};
```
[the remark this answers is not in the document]
TDemeco: No such thing exists, the intended way to call RPC methods using PAPI is the method used.

test/cli/handlers/launch/relayer.ts
```diff
@@ -52,9 +96,43 @@
...
+    substrateNodeId = "default (assumed)";
+  } else {
+    const firstDhNode = dhNodes[0];
+    substrateWsPort = firstDhNode.port;
+    substrateNodeId = firstDhNode.id;
+    logger.info(
+      `🔌 Using DataHaven node ${substrateNodeId} on port ${substrateWsPort} for relayers and BEEFY check.`
+    );
+  }
+
```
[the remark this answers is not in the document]
TDemeco: Defaulting is done to allow relayers to launch over an already launched DataHaven network, since in that case `launchedNetwork` would not have DataHaven nodes.

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@
...
+        logger.success(
+          `Successfully fetched ${authorityPublicKeys.length} BEEFY next authorities directly.`
+        );
+      } else {
+        logger.warn(
+          "⚠️ Fetched BEEFY nextAuthorities is empty. Falling back to hardcoded authority set."
+        );
+        authorityPublicKeys = Object.values(FALLBACK_DATAHAVEN_AUTHORITY_PUBLIC_KEYS);
+      }
+      await papiClient.destroy();
```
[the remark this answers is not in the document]
TDemeco: This is just wrong. The return from the query is an array of `FixedSizeBinary<33>` which represent the public authority keys. This type implements the `.asHex()` method to convert those keys to their hex form in a TypeScript string.

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  ferdie: "0x03bc9d0ca094bd5b8b3225d7651eac5d18c1c04bf8ae8f8b263eebca4e1410ed0c"
+};
+
+// Function to convert compressed public key to Ethereum address
+function compressedPubKeyToEthereumAddress(compressedPubKey: string): string {
+  const uncompressedPubKey = ethers.SigningKey.computePublicKey(compressedPubKey, false);
+  const address = ethers.computeAddress(uncompressedPubKey);
+  return address;
+}
+
```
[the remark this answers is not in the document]
TDemeco: That would still throw, and throwing here is acceptable since it means there's something really wrong with our BEEFY configuration in the runtime.

test/package.json
```diff
@@ -45,6 +47,7 @@
     "commander": "^13.1.0",
     "dockerode": "^4.0.6",
     "dotenv": "^16.5.0",
+    "ethers": "^6.14.0",
```
ffarall: What are you using ethers for? We have viem for that.

test/package.json
```diff
@@ -13,8 +13,10 @@
...
     "generate:types": "(cd ../operator && cargo build --release) && bun x papi add --wasm \"../operator/target/release/wbuild/datahaven-stagenet-runtime/datahaven_stagenet_runtime.wasm\" datahaven",
     "start:e2e:verified": "bun cli --verified --blockscout --deploy-contracts --setup-validators --update-validator-set --fund-validators --slot-time 1",
+    "start:e2e:verified:relayer": "bun cli --verified --blockscout --deploy-contracts --setup-validators --update-validator-set --fund-validators --slot-time 1 --relayer --datahaven",
     "start:e2e:ci": "bun cli -d --setup-validators --update-validator-set --fund-validators --always-clean --slot-time 2 --datahaven --relayer",
-    "start:e2e:minrelayer": "bun cli --relayer -d --no-setup-validators --no-update-validator-set --no-fund-validators --datahaven",
+    "start:e2e:minimal:relayer": "bun cli --relayer -d --no-setup-validators --no-update-validator-set --no-fund-validators --datahaven",
+    "start:e2e:minimal:relayer:fast": "bun cli --relayer -d --no-setup-validators --no-update-validator-set --no-fund-validators --datahaven --fast-runtime",
```
ffarall: This flags will change with my PR getting merged before this.

test/utils/index.ts
```diff
@@ -10,3 +10,4 @@ export * from "./parser";
 export * from "./rpc";
 export * from "./shell";
 export * from "./viem";
+export * from "./kurtosis";
```
ffarall: Leftover?

test/cli/handlers/launch/relayer.ts
```diff
@@ -52,9 +96,43 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
...
+  );
+  const beefyIsReady = await isBeefyReady(substrateWsPort);
+  if (!beefyIsReady) {
+    throw new Error(
+      `BEEFY protocol not ready on port ${substrateWsPort} (node: ${substrateNodeId}). Relayers cannot be launched.`
+    );
+  }
+  logger.success(
+    `🥩 BEEFY is ready (node: ${substrateNodeId}, port: ${substrateWsPort}), proceeding with relayer launch.`
+  );
```
ffarall: Kind of a nitpick, but all these verbose comments (which I support) are probably better suited inside of `isBeefyReady`

test/cli/handlers/launch/relayer.ts
```diff
@@ -17,6 +20,47 @@ import {
 import type { LaunchOptions } from ".";
 import type { LaunchedNetwork } from "./launchedNetwork";
 
+const ZERO_HASH = "0x0000000000000000000000000000000000000000000000000000000000000000";
+
+export const isBeefyReady = async (port: number, retries = 30, delay = 2000): Promise<boolean> => {
```
ffarall: I'd move this to the bottom of hte file, to match the structure of the changes being merged in my PR.

Also, we should have a common "interface" for these polling functions. Mine asks for `period` and `timeout` and calculates the number of retries with that. I'm fine with whichever option, but better to have them more or less the same.

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  charlie: "0x031d10105e323c4afce225208f71a6441ee327a65b9e646e772500c74d31f669aa",
+  dave: "0x0291f1217d5a04cb83312ee3d88a6e6b33284e053e6ccfc3a90339a0299d12967c",
+  eve: "0x0389411795514af1627765eceffcbd002719f031604fadd7d188e2dc585b4e1afb",
+  ferdie: "0x03bc9d0ca094bd5b8b3225d7651eac5d18c1c04bf8ae8f8b263eebca4e1410ed0c"
+};
+
+// Function to convert compressed public key to Ethereum address
+function compressedPubKeyToEthereumAddress(compressedPubKey: string): string {
+  const uncompressedPubKey = ethers.SigningKey.computePublicKey(compressedPubKey, false);
+  const address = ethers.computeAddress(uncompressedPubKey);
```
ffarall: I see that you use it here. Doesn't `viem` have a similar functionality?

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  return address;
+}
+
+/**
+ * Prepares the configuration for Datahaven authorities by converting their
+ * compressed public keys to Ethereum addresses and saving them to a JSON file.
+ */
+export async function setupDatahavenValidatorConfig(
+  options: LaunchOptions,
```
ffarall: not used.

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  const uncompressedPubKey = ethers.SigningKey.computePublicKey(compressedPubKey, false);
+  const address = ethers.computeAddress(uncompressedPubKey);
+  return address;
+}
+
+/**
+ * Prepares the configuration for Datahaven authorities by converting their
+ * compressed public keys to Ethereum addresses and saving them to a JSON file.
+ */
+export async function setupDatahavenValidatorConfig(
```
ffarall:
```suggestion
export async function setupDataHavenValidatorConfig(
```

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+
+// Function to convert compressed public key to Ethereum address
+function compressedPubKeyToEthereumAddress(compressedPubKey: string): string {
+  const uncompressedPubKey = ethers.SigningKey.computePublicKey(compressedPubKey, false);
+  const address = ethers.computeAddress(uncompressedPubKey);
+  return address;
+}
+
+/**
+ * Prepares the configuration for Datahaven authorities by converting their
```
ffarall:
```suggestion
 * Prepares the configuration for DataHaven authorities by converting their
```

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  "--no-telemetry",
+  "--enable-offchain-indexing=true"
 ];
 
 // We need 5 since the (2/3 + 1) of 6 authority set is 5
 // <repo_root>/operator/runtime/src/genesis_config_presets.rs#L94
-const AUTHORITY_IDS = ["alice", "bob", "charlie", "dave", "eve"];
+const CLI_AUTHORITY_IDS = ["alice", "bob", "charlie", "dave", "eve"];
+
+// 33-byte compressed public keys for Datahaven next validator set
```
ffarall:
```suggestion
// 33-byte compressed public keys for DataHaven next validator set
```

test/cli/handlers/launch/datahaven.ts
```diff
@@ -13,12 +18,144 @@ const COMMON_LAUNCH_ARGS = [
...
+  alice: "0x020a1091341fe5664bfa1782d5e04779689068c916b04cb365ec3153755684d9a1",
+  bob: "0x0390084fdbf27d2b79d26a4f13f0ccd982cb755a661969143c37cbc49ef5b91f27",
+  charlie: "0x031d10105e323c4afce225208f71a6441ee327a65b9e646e772500c74d31f669aa",
+  dave: "0x0291f1217d5a04cb83312ee3d88a6e6b33284e053e6ccfc3a90339a0299d12967c",
+  eve: "0x0389411795514af1627765eceffcbd002719f031604fadd7d188e2dc585b4e1afb",
+  ferdie: "0x03bc9d0ca094bd5b8b3225d7651eac5d18c1c04bf8ae8f8b263eebca4e1410ed0c"
+};
+
+// Function to convert compressed public key to Ethereum address
+function compressedPubKeyToEthereumAddress(compressedPubKey: string): string {
```
ffarall: I'd prefer to move these "utility" functions down to the bottom. Also, we're using the `const foo =  () => {}` syntax for functions.

test/package.json
TDemeco: I missed that we already had `viem` as a dependency! Removed `ethers`

test/package.json
TDemeco: Ok! I'll update them

test/utils/index.ts
TDemeco: Yeah, removed

test/cli/handlers/launch/relayer.ts
TDemeco: Refactored! Now we only call `isBeefyReady` and logs are emitted from inside it

test/cli/handlers/launch/relayer.ts
TDemeco: I'll copy your way of doing it since your PR is going to get merged first 😄

test/cli/handlers/launch/datahaven.ts
TDemeco: A little bit more complex but yeah, same can be achieved with `viem`. I already refactored this so feel free to check the new logic out

test/cli/handlers/launch/datahaven.ts
TDemeco: I updated everything that was written as `Datahaven` to `DataHaven` 😬

PR #68, opened 2025-05-12
feat: ✨ Add Polkadot API support to CLI and e2e testing infra

ffarall:
In this PR:
1. Add [Polkadot API](https://papi.how/) support to `test/` directory (i.e. the e2e CLI and testing framework) to be able to interact with the Substrate chain.
    1. This allows typed interactions with transactions, constants, storage, runtime APIs, and non-typed interaction with RPC methods.
    2. Types are autogenerated when running `bun i`, from the `datahaven.scale` file that is part of this repo's version control.
2. Add new utilities file to `papi` related functionalities. For the time being, generating a new signer from a private key.
3. ~Add a new step to the CLI that sends a transaction to the DataHaven network. _*THIS SHOULD BE REMOVED SOON, IT'S JUST FOR TESTING PURPOSES*_~
    1. Both steps that send test transactions have been removed from the CLI, for convenience and being error prone. Their scripts remain usable for testing purposes if needed.
4. Removes the `apis.rs` files from the runtime definitions. Having them in a separate file meant that the runtime was not including the Runtime APIs in the metadata blob, preventing `papi` from creating types from it. This change can be reapplied after upgrading to `polkadot-sdk` `stable-2503`. More info [here](https://github.com/paritytech/polkadot-sdk/issues/6659).
5. Add script to re-generate types, and corresponding docs.
6. Makes logger synchronous to avoid prints happening before logs.

### DIFF
--- a/operator/node/src/consensus.rs
+++ b/operator/node/src/consensus.rs
@@ -3,6 +3,7 @@ use sp_inherents::InherentData;
 use sp_runtime::{generic::Digest, traits::Block as BlockT, DigestItem};
 
 /// Implement pending consensus data provider for BABE.
+#[derive(Default)]
 pub struct BabeConsensusDataProvider {}
 
 impl BabeConsensusDataProvider {
--- a/operator/node/src/service.rs
+++ b/operator/node/src/service.rs
@@ -325,7 +325,7 @@ where
         import_queue,
         keystore_container,
         select_chain,
-        transaction_pool: transaction_pool.into(),
+        transaction_pool,
         other: (
             block_import,
             grandpa_link,
@@ -504,7 +504,7 @@ where
             frontier_partial_components: FrontierPartialComponents {
                 filter_pool: filter_pool.clone(),
                 fee_history_cache: fee_history_cache.clone(),
-                fee_history_cache_limit: fee_history_cache_limit.clone(),
+                fee_history_cache_limit,
             },
             storage_override,
             sync: sync_service.clone(),
--- a/operator/runtime/mainnet/src/configs/mod.rs
+++ b/operator/runtime/mainnet/src/configs/mod.rs
@@ -629,13 +629,13 @@ impl pallet_evm_chain_id::Config for Runtime {}
 
 // --- Snowbridge Config Constants & Parameter Types ---
 parameter_types! {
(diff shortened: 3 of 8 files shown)


test/README.md
```diff
@@ -26,7 +31,7 @@ Follow these steps to set up and interact with your test environment:
 1. **Deploy a minimal test environment**
```
TDemeco, line 31: From here down we should update the docs since they became outdated after this change

test/package.json
```diff
@@ -11,16 +11,17 @@
     "build:docker:relayer": "bun -e \"import build  from './scripts/snowbridge-relayer.ts'; build()\"",
     "generate:wagmi": "wagmi generate",
     "generate:snowbridge-cfgs": "bun -e \"import {generateSnowbridgeConfigs} from './scripts/gen-snowbridge-cfgs.ts'; await generateSnowbridgeConfigs()\"",
+    "generate:types": "(cd ../operator && cargo build --release) && bun x papi add --wasm \"../operator/target/release/wbuild/datahaven-stagenet-runtime/datahaven_stagenet_runtime.wasm\" datahaven",
     "start:e2e:verified": "bun cli --verified --blockscout --deploy-contracts --setup-validators --update-validator-set --fund-validators --slot-time 1",
     "start:e2e:ci": "bun cli -d --setup-validators --update-validator-set --fund-validators --always-clean --slot-time 2 --datahaven --relayer",
     "start:e2e:minrelayer": "bun cli --relayer -d --no-setup-validators --no-update-validator-set --no-fund-validators --datahaven",
     "stop:e2e": "pkill datahaven ; pkill snowbridge-relay ; kurtosis enclave stop datahaven-ethereum && kurtosis clean && kurtosis engine stop && docker container prune -f",
     "stop:e2e:verified": "bun stop:e2e",
-    "stop:e2e:minimal": "bun stop:e2e",
```
TDemeco, line 19: Why delete only `stop:e2e:minimal` and not `stop:e2e:verified` as well?

ffarall: Cause we still have a `start:e2e:verified`. I agree that there's not much use anyways.

test/README.md
ffarall: Agreed. I intend on updating it properly in the other PR where I change this flow further with the relayers functionalities.

PR #69, opened 2025-05-13
DataHaven AVS subgraph initial implementation

stiiifff:


subgraph/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      ethereum: "local:http://host.docker.internal:52157"
+      GRAPH_LOG: info
+  ipfs:
+    image: ipfs/kubo:v0.17.0
+    ports:
+      - "5001:5001"
+    volumes:
+      - ./data/ipfs:/data/ipfs
+  postgres:
+    image: postgres:14
```
undercover-cactus, line 31: The most recent image is postgres:17. We could use this one if there is no particular reason to use 14.

https://hub.docker.com/_/postgres

PR #73, opened 2025-05-15
feat: 🏗️  run execution relayer 

gonzamontiel:
## This PR includes:
- Running the execution relayer on the CLI
- Modifying the Payload generation in the `DataHavenServiceManager.sol`
- Modified the `EigenLayerMessageProcessor` to work with the ValidatorSet update message, but for this change we are loosing the generic message type (it was the only way to make it work so far).
- Adds a `--no-wait` argument to the cli launch and stop commands to bootstrap faster.

### Testing the Snowbridge message encoding / decoding
- Added`MessageEncoding.t.sol` is documented and explains how to generate bin data to use in the rust test.
- Added a Rust unit test to `EigenLayerMessageProcessor` to compare the message encoding/decoding taking the bytes from a file (previously generated with some mock data).

Specifically, we want that:

https://github.com/Moonsong-Labs/datahaven/blob/3cbca0db6d0bcc6cd6969b13b214a27465460b45/contracts/src/libraries/DataHavenSnowbridgeMessages.sol#L78-L85

Generates the right bytes encoding for https://github.com/Moonsong-Labs/datahaven/blob/0e2c9cd518ada0893f1759eee229976fb5a90495/operator/primitives/bridge/src/lib.rs#L51

If the test passes, it's very likely that the CLI will also pass, if not, then we might wanna check something else is missing.

### Breaking change ⚠️ 
For compatibility reasons with Snowbridge contracts (they call specific extrinsics of specific pallets), I had to rename:
- `InboundQueueV2` -> `EthereumInboundQueueV2`
- `OutboundQueueV2` -> `EthereumOutboundQueueV2`

## For follow up PRs:

- Add an automated way of generating the Solidity bytes fo testing, so we don't need to maintain the MessageEncoding.t.sol and generate the binary data manually.

## Summary by CodeRabbit

- **New Features**
  - Added support for the "execution" relayer type in relay configuration, parsing, and CLI launch utilities.
  - Introduced a Solidity test contract for encoding and logging validator set messages.
  - Added a comprehensive "start:all" script to streamline launching and setup processes.

- **Enhancements**
  - Improved message encoding for validator set updates, aligning with new struct field names and message formats.
  - Updated relay configuration schema and validation to support execution relayers and OFAC settings.
  - Increased beacon datastore capacity and adjusted relay scheduling parameters in configuration files.

- **Refactor**
  - Renamed runtime type aliases for inbound/outbound queues to more descriptive names across mainnet, stagenet, and testnet.
  - Centralized and streamlined validator set update logic in CLI utilities.
  - Centralized message decoding logic and improved visibility of message fields in Rust components.

- **Bug Fixes**
  - Improved error handling and decoding logic for message processing in Rust components.

- **Tests**
  - Added Rust and Solidity tests for message encoding and processing validation.

- **Chores**
  - Updated dependencies and feature flags in Rust project configuration.

### DIFF
--- a/test/utils/parser.ts
+++ b/test/utils/parser.ts
@@ -147,11 +147,14 @@ export const ExecutionRelayConfigSchema = z.object({
     totalRelayerCount: z.number(),
     sleepInterval: z.number()
   }),
-  ofac: z.object({
-    enabled: z.boolean(),
-    apiKey: z.string()
-  })
+  ofac: z
+    .object({
+      enabled: z.boolean(),
+      apiKey: z.string()
+    })
+    .optional()
 });
+
 export type ExecutionRelayConfig = z.infer<typeof ExecutionRelayConfigSchema>;
 
 export type RelayerType = "beefy" | "beacon" | "solochain" | "execution";


contracts/src/libraries/DataHavenSnowbridgeMessages.sol
```diff
@@ -57,13 +64,24 @@ library DataHavenSnowbridgeMessages {
...
-            bytes.concat(ScaleCodec.checkedEncodeCompactU32(payload.newValidatorSet.length), accum);
+
+        return bytes.concat(
+            bytes4(0x70150038),
+            bytes1(uint8(Message.V0)),
+            bytes1(uint8(OutboundCommandV1.ReceiveValidators)),
+            ScaleCodec.encodeCompactU32(validatorsLen),
+            validatorsFlattened,
+            ScaleCodec.encodeU64(uint64(epoch))
+        );
```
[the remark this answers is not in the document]
gonzamontiel: The enconding is working properly, we are decoding 64bits. But do you know @ahmadkaouk if this was intented to be 48 bytes ?

test/utils/parser.ts
```diff
@@ -80,24 +126,45 @@ function parseBeefyConfig(config: unknown): BeefyRelayConfig {
...
+
 /**
  * Type Guard to check if a config object is a BeaconRelayConfig
  */
 export function isBeaconConfig(
-  config: BeaconRelayConfig | BeefyRelayConfig
+  config: BeaconRelayConfig | BeefyRelayConfig | ExecutionRelayConfig
 ): config is BeaconRelayConfig {
   return "beacon" in config.source;
 }
```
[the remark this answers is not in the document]
gonzamontiel: We don't really use this function anywhere, are we fine with deleting it @timbrinded  ?

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
```diff
@@ -57,13 +64,23 @@ library DataHavenSnowbridgeMessages {
     function scaleEncodeNewValidatorSetMessagePayload(
         NewValidatorSetPayload memory payload
     ) public pure returns (bytes memory) {
-        // Encode all fields into a buffer.
-        bytes memory accum = hex"";
-        for (uint256 i = 0; i < payload.newValidatorSet.length; i++) {
-            accum = bytes.concat(accum, payload.newValidatorSet[i]);
+        uint32 validatorsLen = uint32(payload.validators.length);
+        bytes32[] memory validatorSet = payload.validators;
+        uint48 epoch = 0;
```
ffarall, line 74: Isn't this hardcoding wrong? Like, it would only work for the first epoch, wouldn't it? If it's not worth it to fix it for this PR, we should at least leave a TODO comment.

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
```diff
@@ -57,13 +64,23 @@ library DataHavenSnowbridgeMessages {
...
+        for (uint32 i = 0; i < validatorSet.length; i++) {
+            validatorsFlattened =
+                bytes.concat(validatorsFlattened, abi.encodePacked(validatorSet[i]));
         }
-        // Encode number of validator addresses, followed by encoded validator addresses.
-        return
-            bytes.concat(ScaleCodec.checkedEncodeCompactU32(payload.newValidatorSet.length), accum);
+
+        return bytes.concat(
+            bytes4(0x70150038),
```
ffarall: What is this first set of bytes? The compact encoded length of the rest of the encoding? I'd prefer a more programmatic way of calculating it, unless we're absolutely sure it shouldn't change. However, if I'm correct on what I think this is, it will change if, for instance, the size of the validator set changes.

contracts/test/MessageEncoding.t.sol
```diff
@@ -0,0 +1,44 @@
...
+pragma solidity ^0.8.27;
+
+import {Test} from "forge-std/Test.sol";
+import {console} from "forge-std/console.sol";
+import {DataHavenSnowbridgeMessages} from "../src/libraries/DataHavenSnowbridgeMessages.sol";
+
+// This test is used to encode the receive validators message and log the hex string.
+// The hex string is then used to generate the .bin file for the Rust test.
+// To generate the .bin file, run:
+// forge test --match-test testEncodeReceiveValidatorsMessageAndLog -vvvvv --fork-url https://rpc.ankr.com/eth_sepolia
```
ffarall: Is the `--fork-url` flag needed? For what?

operator/primitives/bridge/src/lib.rs
```diff
@@ -69,26 +59,116 @@ where
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*; // Imports NewValidatorSet, NewValidatorSetPayload, EL_MESSAGE_ID from parent module
+    use parity_scale_codec::DecodeAll;
+    use std::fs::File;
+    use std::io::Read;
+    use std::path::Path;
```
ffarall: Merge imports

operator/primitives/bridge/src/lib.rs
```diff
@@ -69,26 +59,116 @@ where
...
+    // These constants should match the mock data used in contracts/test/MessageEncoding.t.sol
+    // Follow the instructions in contracts/test/MessageEncoding.t.sol to generate the .bin file.
+
+    const MOCK_NONCE: u64 = 12345u64;
+    const MOCK_TOPIC_HEX: &str = "123456789012345678901234567890123456789012345678901234567890abcd";
+    const MOCK_VALIDATORS_HEX: [&str; 2] = [
+        "0000000000000000000000000000000000000000000000000000000000000001",
+        "0000000000000000000000000000000000000000000000000000000000000002",
+    ];
+    const MOCK_EXTERNAL_INDEX: u64 = 0u64; // Matchs with epoch in Solidity
```
ffarall:
```suggestion
    const MOCK_EXTERNAL_INDEX: u64 = 0u64; // Matches with epoch in Solidity
```

operator/primitives/bridge/src/lib.rs
```diff
@@ -50,9 +40,9 @@ where
                 network: _,
             } => return false,
         };
-        let decode_result = Payload::<T>::decode_all(&mut payload.as_slice());
-        if let Ok(payload) = decode_result {
-            payload.message_id == EL_MESSAGE_ID
+        let decode_result = NewValidatorSet::decode_all(&mut payload.as_slice());
```
ffarall: This is what you said was hardcoded right? Why couldn't we make it decode well with the abstracted `Payload` type? What we should be doing is adjusting the message format on ethereum side to match the abstracted one here 🤔

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
ffarall: OH! It's the "ID" set in the rust code. Well that we should document 😅. Otherwise it's super arbitrary. Also, maybe we can set it to something still arbitrary, but less random.

test/configs/snowbridge/execution-relay.json
```diff
@@ -1,41 +1,44 @@
...
+  },
+  "instantVerification": false,
+  "schedule": {
+    "id": null,
+    "totalRelayerCount": 1,
+    "sleepInterval": 20
+  },
+  "ofac": {
+    "enabled": false,
+    "apiKey": ""
```
ffarall: What's this?

test/configs/snowbridge/execution-relay.json
```diff
@@ -1,41 +1,44 @@
...
+      "Gateway": ""
     },
-    "instantVerification": false,
-    "schedule": {
-        "id": 1,
-        "totalRelayerCount": 1,
-        "sleepInterval": 1
+    "beacon": {
+      "endpoint": "http://127.0.0.1:59807",
+      "stateEndpoint": "http://127.0.0.1:59807",
```
ffarall: We may want to leave these ones as the defaults port used before. We're setting them programmatically anyway in the CLI, when creating the actual config file used.

test/scripts/set-datahaven-parameters.ts
```diff
@@ -42,7 +43,7 @@ export const setDataHavenParameters = async (
     shouldSetParameters = await confirmWithTimeout(
       "Do you want to set the DataHaven runtime parameters?",
       true,
-      10
+      options.wait ? 10 : 0
```
ffarall: This feature is kinda redundant with the flags, isn't it?

ffarall: And in a weird way 😂

test/cli/index.ts
```diff
@@ -74,6 +74,8 @@ program
     "Tag of the relayer",
     "moonsonglabs/snowbridge-relayer:latest"
   )
+  .option("--w, --wait", "Wait for user confirmation on each step", true)
+  .option("--nw, --no-wait", "Skip waiting for user confirmation on each step")
```
ffarall: Instead of this, which seems like a workaround to the interactive feature, but not using the flags, so it overlaps with that, how about having an `--A` flag that enables all the flags. Essentially you should get the same result, but you re-use the already established flags mechanism.

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
gonzamontiel: Ups, I was planning to document this. I don't know who or why this value was decided, but is definitely the rust bytes that work as a MessageId. But you're totally right, I updated the code to be more self documented.

contracts/test/MessageEncoding.t.sol
gonzamontiel: Not needed, updated.

operator/primitives/bridge/src/lib.rs
gonzamontiel: you're absolutely right, I just rollbacked that change to call the generic `Payload::<T>::decode_all()`

ahmadkaouk: Have you tried the approach Tanssi uses on the Ethereum side [here](https://github.com/moondance-labs/tanssi-bridge-relayer/blob/main/overridden_contracts/src/libraries/OSubstrateTypes.sol)? I used the same Rust types as they are for the message and the payload, so their method on the Ethereum side should work as well. The only difference to watch for is that we’re using AccountId20 for validator keys, whereas they use AccountId32.

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
ahmadkaouk: For context, a message ID is required to differentiate between different types of messages. The reason is that scale encoding is not type-aware; the same set of bytes can be decoded into different data structures if their total size matches. The message ID is checked after decoding to ensure the message has been correctly interpreted as intended by the sender and is suitable for processing by this processor. 

The value used is the same as the one Tanssi used — I simply copied it. However, I agree that we can define a better, deterministic way to generate unique IDs for each message type we define.

test/configs/snowbridge/execution-relay.json
gonzamontiel: A configuration for assets control. This is a config for compliance in the US afaik, and was removed from the relayer in my PR: https://github.com/Moonsong-Labs/snowbridge/pull/16, the same they did in Tanssi. 
All that to say, we can remove it.

test/cli/index.ts
gonzamontiel: Done.

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
gonzamontiel:
```suggestion
        // TODO: propagate the correct epoch value from the caller of this function
        uint48 epoch = 0;
```

gonzamontiel: Yes, we could leave it and I can fix it in a follow up PR, but definitely shouldn't be hardcoded. Let's discuss, if agreed we can commit my suggestion.

test/scripts/set-datahaven-parameters.ts
gonzamontiel: Fixed 👌

test/cli/handlers/launch/relayer.ts
```diff
@@ -179,51 +189,60 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
...
+      }
+      case "solochain": {
+        const cfg = parseRelayConfig(json, type);
+        cfg.source.ethereum.endpoint = `ws://host.docker.internal:${ethWsPort}`;
+        cfg.source.solochain.endpoint = `ws://${substrateNodeId}:${substrateWsPort}`;
+        cfg.source.contracts.BeefyClient = beefyClientAddress;
+        cfg.source.contracts.Gateway = gatewayAddress;
+        cfg.source.beacon.endpoint = `http://host.docker.internal:${ethHttpPort}`;
+        cfg.source.beacon.stateEndpoint = `http://host.docker.internal:${ethHttpPort}`;
+        cfg.source.beacon.datastore.location = datastorePath;
```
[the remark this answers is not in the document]
gonzamontiel: @TDemeco can you confirm this review is right for the solochain relayer?

test/configs/snowbridge/execution-relay.json
gonzamontiel: Done

PR #79, opened 2025-05-18
Dockerize DataHaven CLI (wip)

stiiifff:


test/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# install with --production (exclude devDependencies)
+RUN mkdir -p /tmp/prod
+COPY package.json bun.lock /tmp/prod/
+COPY .papi /tmp/prod/.papi
+RUN cd /tmp/prod && bun install --frozen-lockfile --production
+
+# copy node_modules from tmp directory
+# then copy all (non-ignored) project files into the image
+FROM base AS prerelease
+COPY --from=install /tmp/dev/node_modules node_modules
```
timbrinded, line 23: why so granular? Can't we just do `COPY . .`

timbrinded: and rely on dockerignore as a blacklist instead of this whitelist

PR #82, opened 2025-05-21
feat: ✨ Enabling solochain relayer

TDemeco:
This PR:
- Adds launching the new `solochain` relayer in the relayers' script, with its config files and schemas.
- Updates the relayer launching to use switch-case logic since we are now going to be running 4 relayers, making it cleaner.
- Minor CLI fixes (adding additional args to build command for Linux, improving naming, deleting idle containers instead of only active ones).
- Deletes unused files `substrate-relay.json` (now `solochain-relay.json`), `gen-snowbridge-cfgs.ts` and `snowbridge-relayer.ts` (now Snowbridge binary and docker image generation can be done exclusively from our [Snowbridge repo](https://github.com/Moonsong-Labs/snowbridge))
- Updates the `UniversalLocation` of our stagenet runtime to be under the global consensus for XCM instead of actually being the global consensus. This makes it so we can actually use the Snowbridge System pallets to queue up messages. For mainnet we are going to want to have our own Network ID instead of using Polkadot's.

> [!WARNING]
> ~~All in all these changes allows us to run the solochain relayer, but it won't work without the refactor that's being worked on in https://github.com/Moonsong-Labs/snowbridge/pull/14. I'd advise waiting for that PR to be merged before merging this one.~~ MERGED ✅

### DIFF
--- a/test/scripts/set-datahaven-parameters.ts
+++ b/test/scripts/set-datahaven-parameters.ts
@@ -94,6 +94,7 @@ export const setDataHavenParameters = async (
 
   try {
     for (const param of parameters) {
+      // TODO: Add a graceful way to print the value of the parameter, since it won't always be representable as a hex string
       logger.info(`Attempting to set parameter: ${String(param.name)} = ${param.value.asHex()}`);
 
       const setParameterArgs: any = {


test/scripts/snowbridge-relayer.ts
```diff
@@ -5,19 +5,33 @@ import { Octokit } from "octokit";
...
-  if (!fs.existsSync(TMP_DIR)) {
-    fs.mkdirSync(TMP_DIR, { recursive: true });
+const DEFAULT_IMAGE_NAME = "snowbridge-relay:local";
+const RELATIVE_DOCKER_FILE_PATH = "../docker/snowbridge-relayer.dockerfile";
+const CONTEXT = "../.."; // Relative to this script, resolves to project root
+const TMP_DIR = path.resolve(__dirname, "../tmp/bin");
+const TARGET_BINARY_IN_TMP_PATH = path.resolve(TMP_DIR, "snowbridge-relay");
+
+// Parses command line arguments
+function parseArgs() {
```
gonzamontiel: I don't know how you were planning to run this, but since we have a bun command `bun build:docker:relayer`, maybe you want to use https://bun.sh/guides/process/argv

test/scripts/gen-snowbridge-cfgs.ts
```diff
@@ -76,20 +76,30 @@ const executionRelaySchema = z
   })
```
timbrinded: i'm sorry that you put the work into changing this, but i think we dont use this script - so we could use this opportunity to delete it.

Is that right @ffarall ?

test/docker/snowbridge-relayer.dockerfile
```diff
@@ -0,0 +1,18 @@
+# Use a specific Ubuntu version similar to the official image
```
timbrinded: Is this a useful file to include? We already have a way of building a local snowbridge image in the snowbridge repo itself

https://github.com/Moonsong-Labs/snowbridge/blob/solochain/relayer/build-local-docker.sh

test/scripts/snowbridge-relayer.ts
```diff
@@ -5,19 +5,33 @@ import { Octokit } from "octokit";
 import invariant from "tiny-invariant";
```
timbrinded: This file is sadly mostly redundant now that dockerized snowbridge is already part of the `main` branch. We can probably delete this now sorry

test/cli/handlers/launch/relayer.ts
```diff
@@ -164,22 +179,35 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
 
       await Bun.write(outputFilePath, JSON.stringify(cfg, null, 4));
       logger.success(`Updated beacon config written to ${outputFilePath}`);
-    } else {
```
timbrinded, line 175: Now that's its 3 possible values, can we make this a switch / case? We will need to add execution relay type so that will be easier to extend then too

test/cli/handlers/launch/relayer.ts
```diff
@@ -220,6 +248,10 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
         pk.value
       ];
 
+      if (type === "solochain" && secondaryPk) {
+        relayerCommandArgs.push("--substrate.private-key", secondaryPk.value);
```
timbrinded, line 270: Why does solochain require both sets of privatekeys?

test/cli/handlers/launch/relayer.ts
gonzamontiel: I could also do this in https://github.com/Moonsong-Labs/datahaven/pull/73, I was already planning to

test/scripts/gen-snowbridge-cfgs.ts
TDemeco: No worries! If it can be deleted I'll just delete it, I'll wait for Facu just to make sure

test/docker/snowbridge-relayer.dockerfile
TDemeco: Good call. There's one thing that's missing from that script and that is generating the SSZ encodings for Beacon and the type bindings, that we do when running `mage build`. I could add those steps to that script if needed and remove this one

test/scripts/snowbridge-relayer.ts
TDemeco: :'(

TDemeco: Just kidding! That's great actually, less things to maintain

test/cli/handlers/launch/relayer.ts
TDemeco: Good question! It needs an Ethereum private key to relay the messages from the OutboundQueue of DataHaven to the Gateway contract on Eth, and then it needs a Substrate private key to send the proof that it has relayed the aforementioned message, to receive the corresponding payment

test/cli/handlers/launch/relayer.ts
TDemeco: I already did it! Thanks 💯

test/scripts/snowbridge-relayer.ts
TDemeco: Good suggestion! Considering Tim's comment that we might be removing this file entirely I won't change the argument parsing until I know for sure if we actually need this file

test/docker/snowbridge-relayer.dockerfile
timbrinded: Sure, sounds good

operator/runtime/stagenet/src/configs/mod.rs
```diff
@@ -640,7 +640,8 @@ impl pallet_evm_chain_id::Config for Runtime {}
 
 // --- Snowbridge Config Constants & Parameter Types ---
 parameter_types! {
-    pub UniversalLocation: InteriorLocation = Here;
+    pub const ThisNetwork: NetworkId = NetworkId::Polkadot;
```
ffarall, line 644: Add a TODO here to change it

test/cli/handlers/launch/relayer.ts
```diff
@@ -163,31 +178,54 @@ export const launchRelayers = async (options: LaunchOptions, launchedNetwork: La
...
-    } else {
-      const cfg = parseRelayConfig(json, type);
-      cfg.source.polkadot.endpoint = `ws://${substrateNodeId}:${substrateWsPort}`;
-      cfg.sink.ethereum.endpoint = `ws://host.docker.internal:${ethWsPort}`;
-      cfg.sink.contracts.BeefyClient = beefyClientAddress;
-      cfg.sink.contracts.Gateway = gatewayAddress;
-      await Bun.write(outputFilePath, JSON.stringify(cfg, null, 4));
-      logger.success(`Updated beefy config written to ${outputFilePath}`);
+    switch (type) {
+      case "beacon":
```
ffarall, line 182: You'll like the refactor I did for this. Somewhat similar

test/cli/handlers/launch/relayer.ts
```diff
@@ -324,6 +366,7 @@ const waitBeefyReady = async (
  *
  * @param options - Launch options containing the relayer binary path.
  * @param launchedNetwork - An instance of LaunchedNetwork to interact with the running network.
+ * @param datastorePath - The path to the datastore directory.
```
ffarall: Forgot to remove it?

test/scripts/cargo-crossbuild.ts
```diff
@@ -52,7 +52,11 @@ export const cargoCrossbuild = async (options: { datahavenBuildExtraArgs?: strin
 
     const target = "x86_64-unknown-linux-gnu";
     await addRustupTarget(target);
-    const command = `cargo build --target ${target} --release`;
+
+    // Get additional arguments from command line
+    const additionalArgs = options.datahavenBuildExtraArgs ?? "";
```
ffarall, line 57: Good catch

test/scripts/gen-snowbridge-cfgs.ts
ffarall: Yep, doesn't seem to be used. Feel free to delete it.

test/scripts/set-datahaven-parameters.ts
```diff
@@ -94,7 +94,7 @@ export const setDataHavenParameters = async (
 
   try {
     for (const param of parameters) {
-      logger.info(`Attempting to set parameter: ${String(param.name)} = ${String(param.value)}`);
+      logger.info(`Attempting to set parameter: ${String(param.name)} = ${param.value.asHex()}`);
```
ffarall, line 98: I saw that this gets printed as `object` if you don't add that. However, is `asHex` a good way to represent it, for all the kinds of parameters we might set?

test/scripts/snowbridge-relayer.ts
ffarall: Yeah, feel free to delete it. Not planning on supporting relayer binaries again anytime soon.

test/docker/snowbridge-relayer.dockerfile
ffarall: So... does this get removed? I'm with Tim on this one, and also support that you add that encodings step Tobi.

test/scripts/gen-snowbridge-cfgs.ts
TDemeco: Done!

test/docker/snowbridge-relayer.dockerfile
TDemeco: Yep, I'll open a PR on the Snowbridge repo with the changes to the script. I did them on the solochain relayer PR but I reverted them since it was getting a bit messy (before reverting the naming and folder structure)

test/cli/handlers/launch/relayer.ts
TDemeco: Nice! 💯

test/cli/handlers/launch/relayer.ts
TDemeco: Yep :P I was having an error regarding the dockerfile so I added this param to try and fix it, which wasn't actually needed so I removed it but forgot to remove the docs. It's gone now

test/scripts/set-datahaven-parameters.ts
TDemeco: Definitely not, since for example I'm pretty sure it will fail with numeric parameters. I wanted to fix it quickly to get the PR merged and leave fixing this log to the future me once we add more parameters. I added a TODO to make sure I don't forget it (although since it will fail I won't be able to ignore it either way 😬 )

PR #84, opened 2025-05-21
test: ✨ New CLI functions

timbrinded:
## Changes

- New option: `--kurtosis-enclave-name` to allow you to specify a new ethereum network with a different enclave name. Neccesary step for setting up local testing with multiple enclaves running at once
- Refactor: all single dash options (e.g. `-i`) have been renamed to have double dashes `--` for consistency others
- sad times: CLI now must be invoked `bun cli launch` since we have multiple fns now
- Package Update
- New Function: `stop` which allows you to kill all components  (eth, dh, relayers) or only a single part
- Gonza's mac target build fix
- Misc fixes to ci like caching and rate limits

## Additional Comment 

The CLI needs multiple commands and this PR adds the first new one `stop`.

This syntax is faily extensible so @ffarall 's k8 command can follow suite.

Originally we were going to have an `exec` command to expose internal fns to be callable by command line, i.e. generate beacon checkpoint - however the need for it has since been superceded by the deploy to k8 command that's going to be added. I've left the code in place though so when another usecase comes up we can just plug that in.

### DIFF
--- a/test/cli/handlers/consts.ts
+++ b/test/cli/handlers/consts.ts
@@ -1,0 +1,13 @@
+export const DOCKER_NETWORK_NAME = "datahaven-net";
+export const COMPONENTS = {
+  datahaven: {
+    imageName: "moonsonglabs/datahaven",
+    componentName: "Datahaven Network",
+    optionName: "datahaven"
+  },
+  snowbridge: {
+    imageName: "snowbridge-relayer",
+    componentName: "Snowbridge Relayers",
+    optionName: "relayer"
+  }
+} as const;
--- a/test/cli/handlers/index.ts
+++ b/test/cli/handlers/index.ts
@@ -1,3 +1,4 @@
+export * from "./consts";
 export * from "./exec";
 export * from "./launch";
 export * from "./stop";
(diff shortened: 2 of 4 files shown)


test/utils/shell.ts
```diff
@@ -60,7 +60,7 @@ export const runShellCommandWithLogger = async (
 
     await Promise.all([
       readStream(stdoutReader, "stdout", logLevel),
-      readStream(stderrReader, "stderr", "error")
+      readStream(stderrReader, "stderr", logLevel)
```
ffarall: This makes it so that whatever the commands sends through `stderr`, it won't be shown as an error. I know this is a bit of a problem because many commands sometimes send logs through `stderr` that are not necessarily errors. But it also means that some actual error messages might be printed as `DEBUG`, which can be confusing.

timbrinded: The reason why i did this change is because many services print normal logging on the stderr channel so as not to interfere with CLI output which is done on stdout.

in otherwords, if you run `kurtosis engine stop` it wont actually print out anything if you look at stdout only. Same deal with running polkadot-sdk nodes. 

Can change it back to "error" but it rarely gets used for error reporting it appears.

PR #86, opened 2025-05-22
refactor: 🐳 Improve docker caching (again)

timbrinded:
## Changes

- New CI file for making Docker Prod images
- Changed E2E tests use an image built from a local dockerfile
- Some cargo build options to make it quicker
- Fix the cache hit rate
- added `tsgo` preview to the project 😎
  - Can be invoked with `bun tsgo` to typecheck
  - Install in IDE [VSCode](https://code.visualstudio.com/docs/configure/extensions/extension-marketplace) & [Zed](https://github.com/zed-extensions/tsgo) for super-fast inline typechecking (as you type basically)

## Context

This PR attempts to make the frankly unacceptable CI times better. This achieves that aim by making a crappy image for day-to-day usage and let the prod issue take ages since that will be infrequently used. The reason why the original design didn't work for us is because: 1) we are using the free GH runners 2) when we goto baremetal runners we'll lose our rapid caching abilities which make using docker cheap.

Also, we add `tsgo` support to improve devex. The improvement is astounding.

```sh
hyperfine -n tsc "bun tsc --incremental false --extendedDiagnostics" -n tsgo "bun tsgo --incremental false --extendedDiagnostics"
Benchmark 1: tsc
  Time (mean ± σ):      5.500 s ±  0.221 s    [User: 8.939 s, System: 0.400 s]
  Range (min … max):    5.196 s …  5.845 s    10 runs
 
Benchmark 2: tsgo
  Time (mean ± σ):      99.1 ms ±   8.4 ms    [User: 392.8 ms, System: 54.1 ms]
  Range (min … max):    88.3 ms … 116.0 ms    29 runs
 
Summary
  tsgo ran
   55.48 ± 5.22 times faster than tsc
```

### DIFF
--- a/.github/workflows/task-build-operator.yml
+++ b/.github/workflows/task-build-operator.yml
@@ -30,7 +30,9 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 1
-      - uses: ./.github/workflows/actions/cleanup-runner
+          
+      # Re-enable if you get GH issues with the runner being full
+      # - uses: ./.github/workflows/actions/cleanup-runner
           
       - uses: ./.github/workflows/actions/setup-env
         with:


operator/.cargo/config.toml
timbrinded, line 1: @coderabbit what do these all these options actually do? Are they in any way irrelevant or moot given the context of this project?

operator/.cargo/config.toml
```diff
@@ -0,0 +1,13 @@
+[build]
+jobs = 8
```
ffarall: Why limit like this? Won't it slow down compilation in our machines?

.github/workflows/task-docker.yml
```diff
@@ -17,15 +27,20 @@ jobs:
...
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Download binary artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: datahaven-node-${{ inputs.binary-hash }}
+          path: ./operator/target/x86_64-unknown-linux-gnu/release/
 
-      - uses: ./.github/workflows/actions/cleanup-runner
```
ffarall, line 28: No cleaning up any more?

operator/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM docker.io/paritytech/ci-unified:bullseye-1.85.0 AS base
 
 ARG MOLD_VERSION=2.39.0
 ARG SCCACHE_VERSION=0.10.0
-ARG FAST_RUNTIME=TRUE
+ARG FAST_RUNTIME=FALSE
```
ffarall, line 6: If we merge this PR like this, it will break stagenet

ffarall: Actually this is fine

PR #87, opened 2025-05-23
feat: 🚀 Add `deploy` command to CLI

ffarall:
### New Features
1. Add the `deploy` command to our CLI.
    1. Conditionally deploys kurtosis eth network if we're in `stagenet` environment.
    2. Deploys DH nodes.
    3. Deploys contracts (all of them). In `mainnet` and `testnet` it shouldn't deploy EL contracts, but for now that's not implemented.
    4. Configures parameters, validators and relayers in the same way as `launch`.
    5. Currently, it only deploys `beefy` and `beacon` relayers, `execution` and `solochain` relayers are pending for a subsequent PR.
2. Add `waitFor` utility function that receives a lambda.

### Refactors
1. Several common functionalities used both by the `launch` and `deploy` command have been moved to the `cli/handlers/common` directory, from where both commands use them. These include
    1. Checks for installed dependencies.
    2. Common constants.
    3. The `LaunchedNetwork` class has been moved to this directory.
    4. DataHaven nodes common functions.
    5. Kurtosis common functions.
    6. Relayer common functions.
    7. Kubernetes functions (although only used by `deploy`, it seemed fitting to have it here still).
8. Remove CLI questions and separator prints from `deploy-contracts.ts` and `set-datahaven-parameters.ts` scripts. These things should be in the `cli/launch` folder, which consumes the functions in these scripts.
9. Remove `setParametersFromCollection` from `utils` folder and put it in `cli`.
10. Create base snowbridge relayer configs for `local` and `stagenet` as two separate directories.

### Fixes
1. Sets the default time of the `deploy` command to 6s as Lodestar is slower than Lighthouse.
2. In `runShellCommandWithLogger` only print `stderr` if the command fails.

### Additional Minor Changes
1. K8s secret key names changed from `dh-beefy-relay-eth-key` to `dh-beefy-relay-ethereum-key` and `dh-beacon-relay-sub-key` to `dh-beacon-relay-substrate-key`, for simplicity in the deployment script.
11. Update suggested configs for `.vscode` configs.


test/configs/snowbridge/stagenet/solochain-relay.json
```diff
@@ -0,0 +1,49 @@
...
+                "syncCommitteeSize": 512,
+                "slotsInEpoch": 32,
+                "epochsPerSyncCommitteePeriod": 256,
+                "forkVersions": {
+                    "deneb": 0,
+                    "electra": 0
+                }
+            },
+            "datastore": {
+                "location": "/Users/tdemeco/Desktop/Moonsong/datahaven/test/tmp/tobi-test",
```
stiiifff: I'm afraid I don't have that folder on my laptop .. 😅

test/configs/snowbridge/stagenet/beacon-relay.json
```diff
@@ -0,0 +1,29 @@
...
+        "syncCommitteeSize": 512,
+        "slotsInEpoch": 32,
+        "epochsPerSyncCommitteePeriod": 256,
+        "forkVersions": {
+          "deneb": 0,
+          "electra": 0
+        }
+      },
+      "datastore": {
+        "location": "/Users/facundofarall/Desktop/Moonsong/datahaven/test/tmp/facu-test",
```
stiiifff: I'm afraid I don't have that folder on my laptop .. 😅

test/configs/snowbridge/stagenet/solochain-relay.json
ffarall: These get replaced in the CLI, but we could make that more explicit 😂

test/configs/snowbridge/stagenet/beacon-relay.json
ffarall: These get replaced in the CLI, but we could make that more explicit 😂

PR #90, opened 2025-05-28
test: ⚡️ Speed up E2E (part 1)

timbrinded:
## Changes

- New option `--inject-contracts` which relies on the `state-diff.json` file in `/contracts` directory
- new script `StateDiffRecorder.sol` which can be imported to track state diffs
- Inject native balances for validators in eth
- `TestHarness` class: Currently unused (because we've opted for just speeding up E2E), but might be useful in future if we have an integration test where we only start up partial network
- Fixes: 
  - added missing `await`s to latest from main

## Context

This PR aims to speed up network launch times from 6 minutes to less than a minute.

We inject contract state into the genesis block, as well as set balances, in order to reduce time taken on setup. This pattern can be used in future for additional setup as the State recording is in a separate forge script.

## Further Work (not in this PR)

1. Inject token balances for validators (ethereum)
2. Inject validator params (datahaven)
3. Decreasing time taken for first beacon sync (ethereum)

### DIFF
--- a/test/launcher/network/index.ts
+++ b/test/launcher/network/index.ts
@@ -10,7 +10,7 @@ import type { LaunchNetworkResult, NetworkLaunchOptions } from "../types";
 import { LaunchedNetwork } from "../types/launchedNetwork";
 import { checkBaseDependencies } from "../utils";
 import { COMPONENTS } from "../utils/constants";
-import { fundValidators, setupValidators, updateValidatorSet } from "../validators";
+import { fundValidators, updateValidatorSet } from "../validators";
 
 // Authority IDs for test networks
 const TEST_AUTHORITY_IDS = ["alice", "bob"] as const;
@@ -219,7 +219,7 @@ export const launchNetwork = async (
       rpcUrl: launchedNetwork.elRpcUrl
     });
 
-    // We are injecting contracts with already registers validator state 
+    // We are injecting contracts with already registers validator state
     // // 5. Setup validators
     // logger.info("🔐 Setting up validators...");
     // await setupValidators({
--- a/test/scripts/deploy-contracts.ts
+++ b/test/scripts/deploy-contracts.ts
@@ -5,8 +5,7 @@ import {
   logger,
   parseDeploymentsFile,
   parseRewardsInfoFile,
-  runShellCommandWithLogger,
-  waitForNodeToSync
+  runShellCommandWithLogger
 } from "utils";
 import type { ParameterCollection } from "utils/parameters";
 


test/cli/handlers/exec/index.ts
```diff
@@ -1 +1,3 @@
-// TODO
+// TODO: Implement exec handlers
```
undercover-cactus: What is this TODO ? Should it be take care of in this PR ?

test/cli/handlers/deploy/index.ts
```diff
@@ -68,8 +68,8 @@ const deployFunction = async (options: DeployOptions, launchedNetwork: LaunchedN
     rpcUrl: launchedNetwork.elRpcUrl,
     verified: options.verified,
     blockscoutBackendUrl,
-    parameterCollection,
-    skipContracts: options.skipContracts
+    parameterCollection
+    // skipContracts: options.skipContracts
```
undercover-cactus: Uncomment we still need this.

test/launcher/network/index.ts
```diff
@@ -207,20 +208,42 @@ export const launchNetwork = async (
...
     // 5. Setup validators
-    logger.info("🔐 Setting up validators...");
-    await setupValidators({
-      rpcUrl: launchedNetwork.elRpcUrl
-    });
+    // Skipping this because we have injected contracts
+    logger.info("🔐 Setting up validators... (skipping because we have injected contracts)");
+    // await setupValidators({
+    //   rpcUrl: launchedNetwork.elRpcUrl
+    // });
```
undercover-cactus, line 227: Remove commented step.

test/cli/handlers/deploy/contracts.ts
```diff
@@ -1,48 +0,0 @@
...
-  validateDeploymentParams(options);
-
-  // Build contracts
-  await buildContracts();
-
-  // Construct and execute deployment
-  const deployCommand = constructDeployCommand(options);
-  await executeDeployment(deployCommand, options.parameterCollection);
-
-  printDivider();
```
ahmadkaouk, line 49: Is this the function that you removed since it duplicated?

test/scripts/generate-bytecodes.ts
```diff
@@ -0,0 +1,4 @@
+// Script to automate scraping raw bytecodes from a manually deployed network
+// It will create a JSON file with the bytecodes of all contracts, addresses, storages, nonce
+//
+// Append comment to JSON that explains it's autogenerated and how to run this script
```
ahmadkaouk, line 4: Why this script is empty ?

test/cli/handlers/launch/index.ts
```diff
@@ -167,4 +193,7 @@ export const launchPreActionHook = (
   if (deployContracts === false && fundValidators) {
     thisCmd.error("--fundValidators requires --deployContracts to be set");
   }
+  if (injectContracts && !deployContracts) {
+    thisCmd.error("--inject-contracts requires --deploy-contracts to be set");
+  }
 };
```
ahmadkaouk: This check is causing `bun cli launch --all --inject-contracts` to fail although the `--all` enable `deployContracts`

test/cli/handlers/deploy/contracts.ts
undercover-cactus: Yeah that was the function. In the end we need to keep this.

test/scripts/generate-bytecodes.ts
undercover-cactus: No idea.

test/cli/handlers/launch/index.ts
undercover-cactus: I think I found the issue and it is because this part of the code https://github.com/Moonsong-Labs/datahaven/blob/ff6db0917f23795dd6f2796ac78e027ed9a15930/test/cli/handlers/launch/index.ts#L148-L159

Calling `thisCmd.setOptionValue("deployContracts", true);` son't change the value of the variable `deployContracts`. Its value is still `undefined` leading to `!deployContracts` being `true`.

If I put the console.log like this
```javascript
  // If --all is set, enable all components
  if (all) {
    thisCmd.setOptionValue("datahaven", true);
    thisCmd.setOptionValue("buildDatahaven", true);
    thisCmd.setOptionValue("launchKurtosis", true);
    thisCmd.setOptionValue("deployContracts", true);
    thisCmd.setOptionValue("fundValidators", true);
    thisCmd.setOptionValue("setupValidators", true);
    thisCmd.setOptionValue("updateValidatorSet", true);
    thisCmd.setOptionValue("setParameters", true);
    thisCmd.setOptionValue("relayer", true);
    thisCmd.setOptionValue("cleanNetwork", true);
  }

  console.log(datahaven)
  console.log(buildDatahaven)
  console.log(launchKurtosis)
  console.log(deployContracts)
  console.log(fundValidators)
  console.log(updateValidatorSet)
  console.log(setParameters)
  console.log(relayer)
```

I will have printed
```
$ bun run cli/index.ts launch --all --inject-contracts
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
```

test/cli/handlers/exec/index.ts
gonzamontiel: Can we remove this? we can create a ticket for implementing exec handlers if we need them.

test/cli/handlers/launch/index.ts
```diff
@@ -167,4 +193,9 @@ export const launchPreActionHook = (
   if (deployContracts === false && fundValidators) {
     thisCmd.error("--fundValidators requires --deployContracts to be set");
   }
+
```
gonzamontiel, line 218: I think the correct way of checking dynamically the values would be `thisCmd.getOptionValue("deployContracts")`.
Since we are modifying the command with the `--all` option but not the local variables, which are still the user input. I think we need move the checks before setting the `--all`, this way we check user input. If the user has introduced inconsistent arguments, and also `--all`, then the command will fail because of the checks being before.

test/launcher/network/index.ts
```diff
@@ -186,7 +186,8 @@ export const launchNetwork = async (
         kurtosisEnclaveName: kurtosisEnclaveName,
         blockscout: options.blockscout ?? false,
         slotTime: options.slotTime || 2,
-        kurtosisNetworkArgs: options.kurtosisNetworkArgs
+        kurtosisNetworkArgs: options.kurtosisNetworkArgs,
+        injectContracts: true // Forcing it to be true to run e2e tests
```
ahmadkaouk, line 190: I don’t believe we should force the `injectContracts` to true to run e2e tests. This isn’t the right approach because we might need to modify a smart contract during testing. In such cases, we should deploy the updated contract instead of injecting to incorporate our changes.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -97,13 +107,13 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
...
         // Send the new validator set message to the Snowbridge Gateway
         bytes memory message = buildNewValidatorSetMessage();
-        _snowbridgeGateway.v2_sendMessage{value: msg.value}(
-            message,
-            new bytes[](0), // No assets to send
-            bytes(""), // No claimer
-            executionFee,
-            relayerFee
-        );
+        // _snowbridgeGateway.v2_sendMessage{value: msg.value}(
```
gonzamontiel: Why we have this commented?

contracts/src/DataHavenServiceManager.sol
```diff
@@ -51,6 +51,14 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
     /// @inheritdoc IDataHavenServiceManager
     mapping(address => address) public validatorEthAddressToSolochainAddress;
 
+    enum Message {
```
gonzamontiel: why we are redefining this? we have this already in `lib/DataHavenSnowbridgeMessages.sol`

contracts/src/DataHavenServiceManager.sol
```diff
@@ -119,11 +127,34 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
...
+            EL_MESSAGE_ID,
+            bytes1(uint8(Message.V0)),
+            bytes1(uint8(OutboundCommandV1.ReceiveValidators)),
+            ScaleCodec.encodeCompactU32(validatorsLen),
+            validatorsFlattened,
+            ScaleCodec.encodeU64(externalIndex)
+        );
 
         // Return the encoded message
-        return DataHavenSnowbridgeMessages.scaleEncodeNewValidatorSetMessage(newValidatorSetMessage);
```
gonzamontiel, line 126: We cannot merge this. We need to find a way of including DataHavenSnowbridgeMessages, I don't think improving testing times needs to make production code worse.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -119,11 +127,34 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
         }
         DataHavenSnowbridgeMessages.NewValidatorSetPayload memory newValidatorSetPayload =
             DataHavenSnowbridgeMessages.NewValidatorSetPayload({validators: newValidatorSet});
-        DataHavenSnowbridgeMessages.NewValidatorSet memory newValidatorSetMessage =
-            DataHavenSnowbridgeMessages.NewValidatorSet({payload: newValidatorSetPayload});
+
+        // FIXME: temp solution until we figure out why the lib broke when we inject contracts
```
undercover-cactus: We found the reason. This should be removed.

PR #97, opened 2025-06-12
feat(operator):  ✨ Add native token transfer support from Ethereum to DataHaven

ahmadkaouk:
## Overview
This PR implements the **return path** for DataHaven's native token bridge, enabling tokens that were previously transferred from DataHaven to Ethereum to flow back to DataHaven. 

**Context**: DataHaven native tokens can be transferred to Ethereum where they exist as wrapped ERC20 tokens. This PR completes the bidirectional bridge by implementing the mechanism to transfer these tokens back from Ethereum to DataHaven, effectively "unwrapping" them and restoring them to their native form.

## Key Changes
- **Added `NativeTokenTransferMessageProcessor`** , a new message processor specifically designed to handle native token return transfers from Ethereum, and extended `InboundQueueV2` processor tuple to include native token transfer handling alongside existing EigenLayer message processing

#### 1. **Ethereum Initiation**
- **User Transaction**: User calls `v2_sendMessage` on Snowbridge Gateway contract:
  ```solidity
  v2_sendMessage(
    assets: [{ 
      kind: ForeignToken,
      assetId: <registered_native_token_id>,  // Must match pre-registered token ID
      amount: <transfer_amount>
    }],
    claimer: <datahaven_recipient_as_h160>,   // 20-byte recipient address (SCALE-encoded)
    message: [],                              // Empty for native transfers
    // ... standard Snowbridge fees
  )
  ```
- **Token Burning**: Gateway burns the wrapped ERC20 tokens on Ethereum (removing them from circulation)
- **Message Encoding**: Gateway encodes transfer details into standardized Snowbridge message format
- **Event Emission**: `MessageAccepted` event logged on Ethereum for relayers to pick up

#### 2. **Cross-Chain Delivery Infrastructure**
- **Relayer Network**: Snowbridge relayers monitor Ethereum events and submit messages to DataHaven
- **Consensus Verification**: DataHaven's `EthereumBeaconClient` pallet verifies Ethereum beacon chain consensus
- **Cryptographic Proofs**: Messages include Merkle proofs and execution receipts for tamper-proof verification
- **Direct Processing**: Relayers submit messages via `InboundQueueV2::submit()` which processes them immediately 

#### 3. **DataHaven Message Processing**

**Immediate Processing Flow:**
When a relayer calls `InboundQueueV2::submit()`, the message is processed immediately using a tuple-based processor system:

```rust
type MessageProcessor = (
    EigenLayerMessageProcessor<Runtime>,      // Handles validator operations
    NativeTokenTransferMessageProcessor<Runtime>, // Handles native token returns
);
```

**Processing Steps:**
1. **Message Verification**: Cryptographic verification and nonce checking (prevents replay attacks)
2. **Processor Selection**: Evaluation of processors (First processor returning `true` handles the message):
   - `EigenLayerMessageProcessor::can_process_message()` - returns `false` for token transfers
   - `NativeTokenTransferMessageProcessor::can_process_message()` - validates:
     - Native token is registered via `SnowbridgeSystemV2::register_token()` (`T::NativeTokenId::get().is_some()`)
     - All assets are `ForeignTokenERC20` with matching native token ID
     - Assets array is not empty

**Token Unlocking Process:**
1. **Recipient Extraction**: Decodes H160 address from `claimer` field → converts to DataHaven AccountId (must contain valid SCALE-encoded H160 address)
2. **Sovereign Account Transfer**: Moves tokens from Ethereum sovereign account to recipient (requires sufficient balance from previous outbound transfers)
3. **Event Emission**: `TokensUnlocked` event confirms successful transfer completion

### DIFF
--- a/operator/runtime/mainnet/tests/native_token_transfer.rs
+++ b/operator/runtime/mainnet/tests/native_token_transfer.rs
@@ -13,13 +13,13 @@ use datahaven_mainnet_runtime::{
 use dhp_bridge::NativeTokenTransferMessageProcessor;
 use frame_support::{assert_noop, assert_ok, traits::fungible::Inspect};
 use pallet_datahaven_native_transfer::Event as NativeTransferEvent;
-use sp_core::Get;
 use snowbridge_core::TokenIdOf;
 use snowbridge_inbound_queue_primitives::v2::{
     EthereumAsset, Message as SnowbridgeMessage, MessageProcessor, Payload,
 };
 use snowbridge_pallet_outbound_queue_v2::Event as OutboundQueueEvent;
 use snowbridge_pallet_system::NativeToForeignId;
+use sp_core::Get;
 use sp_core::{H160, H256};
 use sp_runtime::DispatchError;
 use xcm::prelude::*;
--- a/operator/runtime/stagenet/tests/native_token_transfer.rs
+++ b/operator/runtime/stagenet/tests/native_token_transfer.rs
@@ -13,13 +13,13 @@ use datahaven_stagenet_runtime::{
 use dhp_bridge::NativeTokenTransferMessageProcessor;
 use frame_support::{assert_noop, assert_ok, traits::fungible::Inspect};
 use pallet_datahaven_native_transfer::Event as NativeTransferEvent;
-use sp_core::Get;
 use snowbridge_core::TokenIdOf;
 use snowbridge_inbound_queue_primitives::v2::{
     EthereumAsset, Message as SnowbridgeMessage, MessageProcessor, Payload,
 };
 use snowbridge_pallet_outbound_queue_v2::Event as OutboundQueueEvent;
 use snowbridge_pallet_system::NativeToForeignId;
+use sp_core::Get;
 use sp_core::{H160, H256};
 use sp_runtime::DispatchError;
 use xcm::prelude::*;
--- a/operator/runtime/testnet/tests/native_token_transfer.rs
+++ b/operator/runtime/testnet/tests/native_token_transfer.rs
@@ -13,13 +13,13 @@ use datahaven_testnet_runtime::{
 use dhp_bridge::NativeTokenTransferMessageProcessor;
 use frame_support::{assert_noop, assert_ok, traits::fungible::Inspect};
 use pallet_datahaven_native_transfer::Event as NativeTransferEvent;
-use sp_core::Get;
 use snowbridge_core::TokenIdOf;
 use snowbridge_inbound_queue_primitives::v2::{
     EthereumAsset, Message as SnowbridgeMessage, MessageProcessor, Payload,
 };
 use snowbridge_pallet_outbound_queue_v2::Event as OutboundQueueEvent;
 use snowbridge_pallet_system::NativeToForeignId;
+use sp_core::Get;
 use sp_core::{H160, H256};
 use sp_runtime::DispatchError;
 use xcm::prelude::*;


operator/runtime/mainnet/tests/native_token_transfer.rs
```diff
@@ -1,423 +1,421 @@
...
-        );
-
-        // Step 6: Verify event emission with all details
-        let expected_event = RuntimeEvent::SnowbridgeSystemV2(SystemV2Event::RegisterToken {
-            location: reanchored_location.clone().into(),
-            foreign_token_id: token_id,
-        });
-        assert_eq!(last_event(), expected_event);
     });
 }
```
TDemeco: Just as a comment, I liked the previous test style where we clearly stated what we were checking using comments. Even if it's obvious for us, it might help someone digging into this code in the future. Not for every test where we do the same of course, because then it just gets repetitive and overly verbose

operator/primitives/bridge/src/lib.rs
```diff
@@ -101,10 +105,100 @@ where
                     validators,
                     external_index,
                 )?;
+                // Build a unique 32-byte identifier using the outbound message nonce
```
TDemeco: Does this have to be unique? From what I've seen this is only used to emit the `MessageReceived` event (although I might be wrong since I haven't deep dived into this) and that event already has the `nonce` as a parameter so also having it as the message ID seems superfluous.
I think it would be more useful as it was before, having a specific message ID dedicated to each message type and using the nonce as the "uniqueness" identifier.

operator/primitives/bridge/src/lib.rs
```diff
@@ -101,10 +105,100 @@ where
...
+
+        // Convert amount to balance type
+        let balance_amount = token_amount
+            .try_into()
+            .map_err(|_| DispatchError::Other("Amount conversion failed"))?;
+
+        // Unlock tokens from the sovereign account
+        pallet_datahaven_native_transfer::Pallet::<T>::unlock_tokens(&recipient, balance_amount)?;
+
+        // Return a 32-byte identifier derived from the message nonce
```
TDemeco: Same comment here^
Maybe something like `VALIDATOR_SET_UPDATE_ID` and `NATIVE_TRANSFER_ID`, something like that. Again, keep in mind I might be missing something and we do need unique IDs 😅

operator/primitives/bridge/src/lib.rs
```diff
@@ -2,7 +2,10 @@
...
 use frame_support::pallet_prelude::*;
 use parity_scale_codec::DecodeAll;
-use snowbridge_inbound_queue_primitives::v2::{Message as SnowbridgeMessage, MessageProcessor};
+use snowbridge_inbound_queue_primitives::v2::{
+    EthereumAsset, Message as SnowbridgeMessage, MessageProcessor,
+};
+use sp_core::H160;
 use sp_std::vec::Vec;
 
 // Message ID. This is not expected to change and its arbitrary bytes defined here.
```
TDemeco, line 11: If we are going to use different message processors for each type of message (which I don't necessarily oppose to), we should refactor the code of the `EigenLayerMessageProcessor` to rename it to `ValidatorSetUpdateMessageProcessor` or something more in-line with what it actually does. Probably for another PR though but it doesn't hurt to add a TODO here!

operator/runtime/mainnet/tests/native_token_transfer.rs
```diff
@@ -1,423 +1,421 @@
...
-                _ => {}
-            }
-        }
+        // Check event was emitted
+        assert!(System::events().iter().any(|e| matches!(
+            &e.event,
+            RuntimeEvent::DataHavenNativeTransfer(
+                NativeTransferEvent::TokensTransferredToEthereum { .. }
+            )
+        )));
```
TDemeco, line 137: While I really like your approach of checking events since it encompasses more situations and guarantees you won't have event "race conditions" when testing (when the event you are expecting is actually being emitted but you don't realize because another one is being emitted afterwards, happened to me in SH and it took me a bit to realize 😅  ), I believe the deleted `last_event()` function that was removed for not being used could have a use case in this specific type of scenario, where we know the event we are expecting is the last one and it should be, otherwise something's wrong. A bit of a nitpick really but wanted to put it out there 😬

operator/runtime/mainnet/tests/common.rs
```diff
@@ -73,10 +76,6 @@ pub fn root_origin() -> RuntimeOrigin {
     RuntimeOrigin::root()
 }
 
-pub fn last_event() -> RuntimeEvent {
-    System::events().pop().expect("Event expected").event
-}
```
TDemeco, line 142: While I don't fully like the current logic of this function (I believe it should just get the `.last()` event instead of popping, but that's just personal preference) I believe it might have some uses, left a comment regarding that above^

operator/runtime/mainnet/tests/native_token_transfer.rs
ahmadkaouk: I've actually opted to remove `last_event` and instead use the more general event matching approach to avoid potential issues (like the one you mentioned) in the future—specifically in cases where the order of emitted events might change or new events are introduced. This way, the test remains resilient and accurate regardless of event order or additions.

operator/runtime/mainnet/tests/native_token_transfer.rs
ahmadkaouk: Oh yes. I might removed some comments while trying to refactor the tests. I also think it's good to have such comments.

operator/primitives/bridge/src/lib.rs
ahmadkaouk: You’re right — it doesn’t have to be unique. This was initially introduced for XCM messages, I believe, and we don't use it. At the time, I didn’t pay close attention to the fact that the event already includes the nonce. I just wanted to ensure we had something unique to use for message_id, mostly to help with debugging.

That said, since the nonce is already present, we can revert to the old way. We just need to  keep in mind that message_id here reflects the type of message and is not unique per message, so it shouldn’t be used as a uniqueness guarantee in the future.

operator/primitives/bridge/src/lib.rs
ahmadkaouk: The intention wasn’t to have a message processor per message type — even though it might seem that way based on the current setup. Each message processor should handle a broader scope, such as a specific domain (e.g., one for the Eigenlayer that processes all related messages, one for token transfer...). We don’t want to end up with a processor per message, since the number of message types will only continue to grow.
I’m still unsure what the best approach is — whether we should go with a single, centralized processor that handles everything, or break things down by functionality.

operator/primitives/bridge/src/lib.rs
ahmadkaouk: Fixed in 5cfc7ea9ffe9466f128848a300b267aaf00e32d1

operator/primitives/bridge/src/lib.rs
ahmadkaouk: Fixed in 5cfc7ea9ffe9466f128848a300b267aaf00e32d1

PR #99, opened 2025-06-12
feat: ✨ set rewards info as parameters in runtime

TDemeco:
This PR improves the CLI to get from the deployments the `RewardsRegistryAddress` (address of the RewardsRegistry contract deployed), `RewardsAgentOrigin` (origin used for the agent in charge of updating the rewards merkle root in the RewardsRegistry contract) and `RewardsUpdateSelector` (function selector of the function that the agent must execute to do the aforementioned update) and then set these values in the `parameters` pallet of the runtime.

After these changes the rewards merkle root is being updated on the Ethereum side. 🎉

### DIFF
--- a/test/scripts/deploy-contracts.ts
+++ b/test/scripts/deploy-contracts.ts
@@ -1,6 +1,11 @@
 import { $ } from "bun";
 import invariant from "tiny-invariant";
-import { logger, parseDeploymentsFile, parseRewardsInfoFile, runShellCommandWithLogger } from "utils";
+import {
+  logger,
+  parseDeploymentsFile,
+  parseRewardsInfoFile,
+  runShellCommandWithLogger
+} from "utils";
 import type { ParameterCollection } from "utils/parameters";
 
 interface ContractDeploymentOptions {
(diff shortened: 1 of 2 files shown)


operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -942,11 +942,11 @@ impl pallet_external_validators_rewards::types::SendMessage for RewardsSendAdapt
         let command = Command::CallContract {
             target: runtime_params::dynamic_params::runtime_config::RewardsRegistryAddress::get(),
             calldata,
-            gas: 100_0000, // TODO: Determine appropriate gas value after testing
+            gas: 1_000_000, // TODO: Determine appropriate gas value after testing
```
gonzamontiel, line 945: hey, why this increase ? I guess we'll change it later and remove the TODO but I was curious why it was needed

TDemeco: It's not an increase, it was just badly formatted before 😆

gonzamontiel: hahaha 🙈

PR #101, opened 2025-06-16
feat: ✨ add Bun version check to CLI dependencies

ahmadkaouk:
Add version validation to ensure Bun 1.2+ is used for consistent behavior across development environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ target/
 cspell.json
 
 tmp/*
+
 # macOS system files
 .DS_Store
 **/.DS_Store
+
+.worktrees/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,87 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+DataHaven is an EVM-compatible Substrate blockchain secured by EigenLayer. It bridges Ethereum and Substrate ecosystems through:
+- EigenLayer AVS integration for security
+- Snowbridge for cross-chain communication
+- Frontier pallets for EVM compatibility
+- External validators with rewards system
+
+## Critical Development Commands
+
+### E2E Testing Environment (from `/test` directory)
+
+```bash
+# Setup
+bun i                               # Install dependencies
+bun cli                            # Interactive CLI for test environment
+
+# Code Quality
+bun fmt:fix                        # Fix TypeScript formatting
+bun typecheck                      # TypeScript type checking
+
+# Code Generation (run after contract changes)
+bun generate:wagmi                 # Generate TypeScript contract bindings
+bun generate:types                 # Generate Polkadot-API types from runtime
+
+# Local Development
+bun build:docker:operator          # Build local DataHaven Docker image
+bun start:e2e:local               # Launch local test network
+bun stop:e2e                      # Stop all test services
+
+# Testing
+bun test:e2e                      # Run E2E test suite
+```
+
+### Rust/Operator Development
+
+```bash
+cd operator
+cargo build --release --features fast-runtime    # Development build
+cargo test                                       # Run tests
+```
+
+### Smart Contracts (from `/contracts` directory)
+
+```bash
+forge build                        # Build contracts
+forge test                         # Run tests
+forge fmt                          # Format Solidity code
+```
+
+## Architecture Essentials
+
+### Cross-Component Dependencies
+- **Contracts → Operator**: DataHaven AVS contracts register operators and manage slashing
+- **Operator → Contracts**: Operator reads validator registry from contracts
+- **Test → Both**: E2E tests deploy contracts and run operator nodes
+- **Snowbridge**: Enables native token transfers and message passing between chains
+
+### Key Design Patterns
+1. **Service Manager Pattern**: Contracts use EigenLayer's service manager for operator coordination
+2. **Rewards Registry**: Tracks validator performance and distributes rewards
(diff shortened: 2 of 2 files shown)


test/cli/handlers/common/checks.ts
```diff
@@ -15,6 +15,13 @@ export const checkBaseDependencies = async (): Promise<void> => {
 
   logger.success("Kurtosis CLI found");
 
+  if (!(await checkBunVersion())) {
+    logger.error("Bun CLI version must be 1.2 or higher. Please upgrade to v1.2 or later.");
```
gonzamontiel:
```suggestion
    logger.error("Your bun version must be 1.2 or higher. Please run: bun upgrade.");
```

test/cli/handlers/common/checks.ts
```diff
@@ -15,6 +15,13 @@ export const checkBaseDependencies = async (): Promise<void> => {
...
+    throw Error("❌ Bun CLI version is too old.");
```
gonzamontiel:
```suggestion
    throw Error("❌ Bun version is too old.");
```

test/cli/handlers/common/checks.ts
```diff
@@ -15,6 +15,13 @@ export const checkBaseDependencies = async (): Promise<void> => {
...
+  }
+
+  logger.success("Bun CLI found");
```
gonzamontiel:
```suggestion
  logger.success("Bun installed and up to date.");
```

test/cli/handlers/common/checks.ts
```diff
@@ -72,6 +79,28 @@ export const deploymentChecks = async (
   printDivider();
 };
 
+const checkBunVersion = async (): Promise<boolean> => {
+  const { exitCode, stderr, stdout } = await $`bun --version`.nothrow().quiet();
```
gonzamontiel: You have available `Bun.version` so you can skip calling the command and checking the errors.

test/cli/handlers/common/checks.ts
```diff
@@ -72,6 +79,28 @@ export const deploymentChecks = async (
...
+  if (exitCode !== 0) {
+    logger.error(stderr.toString());
+    return false;
+  }
+
+  const bunVersion = stdout.toString().trim();
+  const [major, minor] = bunVersion.split(".").map(Number);
```
gonzamontiel, line 87: Like here it would be:

```suggestion
  const [major, minor] = Bun.version.split(".").map(Number);
```

ahmadkaouk: [8a4bc9d](https://github.com/Moonsong-Labs/datahaven/pull/101/commits/8a4bc9d542e0e8b0ce055652d07696199a885dc3)

test/cli/handlers/common/checks.ts
ahmadkaouk: [8a4bc9d](https://github.com/Moonsong-Labs/datahaven/pull/101/commits/8a4bc9d542e0e8b0ce055652d07696199a885dc3)

test/cli/handlers/common/checks.ts
ahmadkaouk: [8a4bc9d](https://github.com/Moonsong-Labs/datahaven/pull/101/commits/8a4bc9d542e0e8b0ce055652d07696199a885dc3)

test/cli/handlers/common/checks.ts
ahmadkaouk: [8a4bc9d](https://github.com/Moonsong-Labs/datahaven/pull/101/commits/8a4bc9d542e0e8b0ce055652d07696199a885dc3)

test/cli/handlers/common/checks.ts
ahmadkaouk: [8a4bc9d](https://github.com/Moonsong-Labs/datahaven/pull/101/commits/8a4bc9d542e0e8b0ce055652d07696199a885dc3)

test/cli/handlers/common/checks.ts
```diff
@@ -15,6 +15,13 @@ export const checkBaseDependencies = async (): Promise<void> => {
 
   logger.success("Kurtosis CLI found");
 
+  if (!(await checkBunVersion())) {
+    logger.error("Bun version must be 1.2 or higher. Run `bun upgrade` to upgrade.");
+    throw Error("❌ Bun version is too old.");
```
ffarall: Running `bun upgrade` won't upgrade it. We should refer them to the docs.

test/cli/handlers/common/checks.ts
```diff
@@ -72,6 +79,22 @@ export const deploymentChecks = async (
   printDivider();
 };
 
+const checkBunVersion = async (): Promise<boolean> => {
+  const bunVersion = Bun.version;
+  const [major, minor] = bunVersion.split(".").map(Number);
+
+  // Check if version meets minimum requirements (1.2+)
+  const isVersionValid = major > 1 || (major === 1 && minor >= 2);
```
ffarall: We can probably put this in the `common/consts.ts` file, to find it later easier and change it.

test/cli/handlers/common/checks.ts
ffarall: Also here, ideally let's not hardcode the version number required to be printed, but let's use the constant.

gonzamontiel: > Running `bun upgrade` won't upgrade it. We should refer them to the docs.

This worked for me at least, it wont go to the desired 1.2 version, but the latest which should also work.
The best way to do it IMO is to have a `.bun-version` file as other projects have adopted, and then we can just enforce that version from the code. But for now, I think we are good to go with this PR.

PR #103, opened 2025-06-17
feat: ✨ Deployment improvements & environmental overrides

stiiifff:
This PR contains improvements to the DataHaven deployment infrastructure:
  1. Directory restructure: Moved from `deployment/` to `deploy/` (more common for K8s / Helm -based deployment configs).
  2. Added **local environment** support: updated CLI to support deploying to a local K8s cluster.
  3. Manual deployment script: `deploy/scripts/deploy.sh` for manual deployments.
  4. Environment-specific configurations: Structured values files for each environment.
  5. Chart organization: Renamed bridges-common-relay to relay for clarity.


test/cli/handlers/deploy/cleanup.ts
```diff
@@ -53,15 +53,18 @@ export const cleanup = async (
...
 
-  // Check if the Kurtosis gateway process is running.
-  const { exitCode, stdout } = await $`pgrep -f "kurtosis gateway"`.nothrow().quiet();
-  if (exitCode !== 0) {
-    logger.error(
-      "❌ `kurtosis gateway` process not found running. This is required for Kurtosis to work with Kubernetes."
-    );
-    throw new Error("Kurtosis gateway process not found running.");
+  // Skip gateway check for local deployments
+  if (options.environment !== "local") {
```
stiiifff: @gonzamontiel Did the whole local deployment succeed without the Kurtosis gateway ? I had it running (for the local K8s / Kurtosis environment) and all was fine.

gonzamontiel: In my case the check was failing because `kurtosis gateway` failed for me when my cluster wasn't set to "cloud". I will need to double check this though.

gonzamontiel: It turns out that `kurtosis gateway` is needed always that the type of the cluster is `kubernetes`, I was using docker directly.

deploy/README.md
```diff
@@ -0,0 +1,208 @@
...
+- **Snowbridge Relays**: Handle cross-chain communication with Ethereum
+  - **Beacon Relay**: Relays Ethereum beacon chain data
+  - **BEEFY Relay**: Relays BEEFY consensus data for finality
+  - **Execution Relay**: Relays Ethereum execution layer data
+
+## Development
+
+For local development:
+1. Ensure you have a local Kubernetes cluster (e.g., Minikube, Kind)
+2. Deploy using the local environment:
```
gonzamontiel, line 100: It is a bit confusing that we have this script and also we have `bun cli deploy`, they do the same? Is this a backup? We should be clear about it

deploy/environments/mainnet/values.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Override settings for bootnode
+dh-bootnode:
+  node:
+    customNodeKey: 80c30ac6ba927c6e5c0c9681aa9674f1d181d180853bcd3485cee9d18e931238
+
+# Override settings for validator
+dh-validator:
+  node:
+    customNodeKey:
+      # To generate new key run: docker run --rm -t moonsonglabs/datahaven:latest key generate-node-key
```
gonzamontiel: Is this intended to be the actual key, or is intended to generate a new one? it's also not clear

deploy/README.md
stiiifff: As mentioned in the PR description: deploy/scripts/deploy.sh for manual deployments.

deploy/environments/mainnet/values.yaml
stiiifff: I will actually remove the testnet & mainnet configs. I am not sure we will use the current K8s setup for those environments.

deploy/README.md
gonzamontiel: I meant in the documentation, but I guess with the ones I added we are good to go!

PR #105, opened 2025-06-18
docs: 📖 add deployment documentation

gonzamontiel:
- [x] Include prerequisites and deployment instructions for deploying to k8s cluster
- [x] Include some helper commands

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -12,7 +12,7 @@
 ## Pre-requisites
 
 - [Kurtosis](https://docs.kurtosis.com/install): For launching test networks
-- [Bun](https://bun.sh/): TypeScript runtime and package manager
+- [Bun](https://bun.sh/) v1.2 or higher: TypeScript runtime and package manager
 - [Docker](https://www.docker.com/): For container management
 
 ##### MacOS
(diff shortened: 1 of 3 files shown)


test/docs/deployment.md
```diff
@@ -0,0 +1,358 @@
...
+   curl -L https://github.com/kurtosis-tech/kurtosis-cli-release-artifacts/releases/latest/download/install-kurtosis.sh | bash
+   
+   # Verify installation
+   kurtosis version
+   ```
+
+3. **Bun** (TypeScript runtime)
+   ```bash
+   # macOS
+   brew install bun
```
ffarall: This is not the way to install bun with brew.

ffarall: See: https://bun.sh/docs/installation

ffarall: In fact, generally speaking, I'd prefer if we point the user to the installation docs of each dependency, rather than giving them the command.

gonzamontiel: Ups. missed that one.
Yeah, I intended in purpose to be a self-contained doc. I know we cannot cover every case, so I tried to cover Mac and Linux. I'll add, on top of that, a link to installation docs for reference.

gonzamontiel:
```suggestion
   brew install oven-sh/bun/bun
```

PR #106, opened 2025-06-18
feat: :sparkles: make `RewardsRegistry` keep a history of roots and claim status

TDemeco:
# Description
This PR implements a comprehensive overhaul of the `RewardsRegistry` contract to maintain complete history of reward merkle roots while providing index-based claim tracking for operators. The new architecture enables operators to claim rewards from any historical merkle root instead of only the latest one. To do so, it:
- Adds the `merkleRootHistory` storage array to the contract, in which we keep all rewards roots that ever came from the DataHaven side.
- Adds the `operatorClaimedByIndex` storage map to the contract, in which we keep track, for each validator and root index, if it has claimed it or not.
  - This works even for new validators, since theoretically with this system you could argue they could claim older roots that they were not a part of which would be catastrophic, but they could never draft a correct proof for those to claim them.
- Keeps some of the interface from before the overhaul, to have quick access to the latest rewards merkle root through `getLatestMerkleRoot()` and to claim rewards for it with `claimRewards()`. This is because the expected behaviour is for validators to claim their rewards every era.
- Adds a way to batch claim rewards with `claimRewardsBatch()`. This function allows a validator to claim rewards for multiple root indices in one call by providing multiple proofs, useful if the validator has fallen behind claims and has to catch up, although special care will have to be taken by it to avoid reaching the gas limit of a transaction.

## Storage Efficiency Analysis
One might think this solution is not as storage-efficient as other solutions that we can think of (I even had two other alternatives in mind as well), but a simple back-of-the-envelope calculation gives us peace of mind that the impact of this solution on the overal state size of the chain is negligible:

### Assumptions (Worst Case Scenario):
- 1,000 validators (actual estimate for DataHaven: ~50/100 validators)
- 6-hour eras (most-likely scenario, following what Polkadot does: ~24-hour eras)
  - Which means 4 merkle root updates per day

### Annual Storage Requirements:
- Merkle Root History: **46,720 bytes/year**
  - 4 roots/day × 32 bytes/root × 365 days/year = 46,720 bytes/year
- Operator Claim Tracking: **~1.46 MB/year**
  - 1,000 operators × 1 boolean/(operator * root index) × 1 byte/boolean × 4 root indices/day × 365 days/year = 1,460,000 bytes/year
- **Total: ~1.5 MB/year**

This represents negligible storage overhead compared to the significant operational benefits gained.

## TODO
Since we want to allow the operators/validators to only have to interact with the AVS contract (that's why the `claimRewards` functions have the `onlyAVS` modifier), we still have to:
- [x] Add the required functions to the AVS to allow operators to claim their rewards.
- [x] Adds comprehensive unit tests for them.

### DIFF
--- a/test/contract-bindings/generated.ts
+++ b/test/contract-bindings/generated.ts
@@ -2105,6 +2105,18 @@ export const dataHavenServiceManagerAbi = [
       { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
       { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
     ],
+    name: 'claimLatestOperatorRewards',
+    outputs: [],
+    stateMutability: 'nonpayable',
+  },
+  {
+    type: 'function',
+    inputs: [
+      { name: 'operatorSetId', internalType: 'uint32', type: 'uint32' },
+      { name: 'rootIndex', internalType: 'uint256', type: 'uint256' },
+      { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
+      { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
+    ],
     name: 'claimOperatorRewards',
     outputs: [],
     stateMutability: 'nonpayable',
@@ -2118,18 +2130,6 @@ export const dataHavenServiceManagerAbi = [
       { name: 'proofs', internalType: 'bytes32[][]', type: 'bytes32[][]' },
     ],
     name: 'claimOperatorRewardsBatch',
-    outputs: [],
-    stateMutability: 'nonpayable',
-  },
-  {
-    type: 'function',
-    inputs: [
-      { name: 'operatorSetId', internalType: 'uint32', type: 'uint32' },
-      { name: 'rootIndex', internalType: 'uint256', type: 'uint256' },
-      { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
-      { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
-    ],
-    name: 'claimOperatorRewardsByIndex',
     outputs: [],
     stateMutability: 'nonpayable',
   },
@@ -8137,6 +8137,18 @@ export const rewardsRegistryAbi = [
       { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
       { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
     ],
+    name: 'claimLatestRewards',
+    outputs: [],
+    stateMutability: 'nonpayable',
+  },
+  {
+    type: 'function',
+    inputs: [
+      { name: 'operatorAddress', internalType: 'address', type: 'address' },
+      { name: 'rootIndex', internalType: 'uint256', type: 'uint256' },
+      { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
+      { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
+    ],
     name: 'claimRewards',
     outputs: [],
     stateMutability: 'nonpayable',
@@ -8150,18 +8162,6 @@ export const rewardsRegistryAbi = [
       { name: 'proofs', internalType: 'bytes32[][]', type: 'bytes32[][]' },
     ],
     name: 'claimRewardsBatch',
-    outputs: [],
-    stateMutability: 'nonpayable',
-  },
-  {
-    type: 'function',
-    inputs: [
-      { name: 'operatorAddress', internalType: 'address', type: 'address' },
-      { name: 'rootIndex', internalType: 'uint256', type: 'uint256' },
-      { name: 'operatorPoints', internalType: 'uint256', type: 'uint256' },
-      { name: 'proof', internalType: 'bytes32[]', type: 'bytes32[]' },
-    ],
-    name: 'claimRewardsByIndex',
     outputs: [],
     stateMutability: 'nonpayable',
   },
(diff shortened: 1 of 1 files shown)


contracts/src/middleware/RewardsRegistryStorage.sol
```diff
@@ -26,11 +26,11 @@ abstract contract RewardsRegistryStorage is IRewardsRegistry {
     /// @notice Address of the rewards agent contract
     address public rewardsAgent;
 
-    /// @notice Last rewards merkle root
-    bytes32 public lastRewardsMerkleRoot;
+    /// @notice History of all merkle roots, accessible by index
+    bytes32[] public merkleRootHistory;
```
ahmadkaouk, line 30: I'm confused about how the "index" in RewardsRegistry relates to eras, since rewards are distributed per era.
Looking a the code, my understanding is that this index is meant to map to era numbers ( era 0 → index 0, era 1 → index 1, etc.).

If this is the case, we have a problem: cross-chain messages in Snowbridge V2 don't guarantee sequential delivery.
```
Example scenario:
Era 0 ends → Snowbridge queues Era 0 merkle root
Era 1 ends → Snowbridge queues Era 1 merkle root
Era 1 message arrives first
Era 0 message arrives second

Result:
merkleRootHistory[0] = Era 1 root ❌ (should be Era 0)
merkleRootHistory[1] = Era 0 root ❌ (should be Era 1)
```
This seems like a design flaw. The contract implicitly assumes sequential message delivery, which isn't  the case with Snowbridge V2.

contracts/src/middleware/RewardsRegistry.sol
```diff
@@ -72,36 +80,171 @@ contract RewardsRegistry is RewardsRegistryStorage {
...
+        for (uint256 i = 0; i < rootIndices.length; i++) {
+            totalRewards +=
+                _performClaim(operatorAddress, rootIndices[i], operatorPoints[i], proofs[i]);
+        }
+
+        // Emit the corresponding event
+        emit RewardsBatchClaimedForIndices(
+            operatorAddress, rootIndices, operatorPoints, totalRewards
+        );
+    }
```
ahmadkaouk: The `claimRewardsBatch`  performs multiple ETH transfers - one for each claim in the batch in the (_performClaim):

https://github.com/Moonsong-Labs/datahaven/blob/805c703f4e01713ff8abf2639eab9ca8c1771803/contracts/src/middleware/RewardsRegistry.sol#L187-L190

This is gas-inefficient. A better solution is to accumulate the total rewards first, then make a single transfer. This would reduce gas costs and simplify error handling (single transfer point).

contracts/src/middleware/RewardsRegistry.sol
```diff
@@ -72,36 +80,171 @@ contract RewardsRegistry is RewardsRegistryStorage {
...
+
+    /**
+     * @notice Claim rewards for an operator from a specific merkle root index
+     * @param operatorAddress Address of the operator to receive rewards
+     * @param rootIndex Index of the merkle root to claim from
+     * @param operatorPoints Points earned by the operator
+     * @param proof Merkle proof to validate the operator's rewards
+     * @dev Only callable by the AVS (Service Manager)
+     */
+    function claimRewardsByIndex(
```
ahmadkaouk: A personal preference, I think that the current function naming creates unnecessary confusion:
```
claimRewards()              // Claims from latest index
claimRewardsByIndex(index)  // Claims from specific index
hasClaimedByIndex(operator, index)
```
The "ByIndex" suffix is misleading because it implies there are alternative ways to claim rewards, when in reality all claims are by index  (either explicitly or implicitly via "latest").

I would suggest the following methods:
```
// Standard function - explicitly specify which root to claim from
claimRewards(uint256 rootIndex)

// Convenience function - special case for latest root
claimLatestRewards()
```

contracts/src/middleware/RewardsRegistry.sol
undercover-cactus: Yep. Good catch.

contracts/src/middleware/RewardsRegistryStorage.sol
TDemeco: That's a good interpretation! Actually, I abstracted the contract from the concept of eras exactly to avoid this issue. There's no connection between indexes and eras, and operators can claim their rewards for any index. This is what I think the flow should look like for an operator trying to claim rewards:
- Check unclaimed indexes.
- Get the root of those unclaimed indexes.
- Check the blocks where those roots (actually the corresponding Snowbridge message) were emitted.
- For those blocks, build the proof for the rewards.
- Submit those proofs alongside the required BEEFY proofs.

Of course, if an operator is on top of its reward claiming, and considering eras will be 24hs long (which alongside the incentives for relayers would mean that in practice the reward root indexes will closely follow the eras indexes) then the process is easier as it would mean:
- After an era end, process and save the required proofs to claim its rewards.
- Once a relayer relays the corresponding message, call the `claimRewards` function of the contract with those proofs.

contracts/src/middleware/RewardsRegistry.sol
TDemeco: Yes! My initial design was actually that way but I ended up changing it to be able to reuse the `_performClaim` function. Ideally we would still reuse an internal function but that instead of actually transferring the funds, it should return the amount to claim (something like `_validateClaim` that returns a `u256` for example) so we can:
- Call that and then transfer the corresponding funds in the single-index claim functions.
- Call that for all root indexes, adding up to the total amount to claim, and then transfer the total amount of funds in the batch claim functions.

contracts/src/middleware/RewardsRegistry.sol
TDemeco: I wanted to preserve the previous interface of the contract but considering we are still a long way from live deployment we can change this up, totally agree.

contracts/src/middleware/RewardsRegistryStorage.sol
ahmadkaouk: Okay, I initially assumed that eras were tightly coupled to indexes. Make sense if we want to change how rewards are computed or distributed in the future. The flow is a bit complex, though. I imagine operators will likely batch their reward claims to optimize for gas costs, so they will need to go with the first approach.

TDemeco: To make it simpler the operators might also, at the end of each era (when the information is fresh) compute the required proofs and save them under the rewards root key in their DB. Then, when batching, it's just a matter of checking the corresponding root of each unclaimed index and getting the already-calculated data from the DB.
We could even add another map to the contract of `rewardsRoot -> rootIndex` to make it even simpler (in that case, the operator would only have to query the new map to get the indexes to pass to the claim function)

ahmadkaouk: I think this is a good first version. To streamline the operator experience, we can provide them with an automated script that handles reward claiming, eliminating the need for manual intervention in this process.

PR #108, opened 2025-06-26
test: 🏗️  small cli fixes

gonzamontiel:
- Added a parameter `--all` to `bun cli launch` (now coherent with `bun cli stop`)
  -  Equivalent to run `bun cli launch --d --bd --lk --dc --fv --sv --uv --sp --r --cn`
- Removed `bun start:all` command 
- Added a kurtosis cluster type check
  - The problem was that we now support deploying and launching the CLI tool with different types of clusters. If you're deploying, your kurtosis cluster most likely needs to point to a `kubernetes` type cluster, if you're running `bun cli launch` you need to use native docker containers. This PR adds a check and warns the user to point to the right config.
  - *Disclaimer*: we don't know the cluster name of the user so we can't force it to be anything 🫠

### DIFF
--- a/test/cli/handlers/common/checks.ts
+++ b/test/cli/handlers/common/checks.ts
@@ -151,7 +151,7 @@ export const checkKurtosisCluster = async (kubernetes?: boolean): Promise<boolea
     return false;
   }
 
-  logger.success(`✅ Kurtosis cluster type "${clusterType}" is compatible`);
+  logger.success(`Kurtosis cluster type "${clusterType}" is compatible`);
   return true;
 };
 


test/cli/handlers/common/checks.ts
```diff
@@ -111,6 +111,50 @@ const checkKurtosisInstalled = async (): Promise<boolean> => {
...
+    logger.error(`❌ Kurtosis cluster type is "${clusterType}" but kubernetes is required`);
+    return false;
+  }
+
+  if (!kubernetes && clusterType !== "docker") {
+    logger.error(`❌ Kurtosis cluster type is "${clusterType}" but docker is required`);
+    return false;
+  }
+
+  logger.success(`✅ Kurtosis cluster type "${clusterType}" is compatible`);
```
ffarall: Using `logger.success` already adds the ✅ emoji automatically. So remove it from here.

gonzamontiel:
```suggestion
  logger.success(`Kurtosis cluster type "${clusterType}" is compatible`);
```

PR #110, opened 2025-06-26
test: init benchmark infra

elfedy:
* Add `run_benchmarks.sh` script to run runtime benchmarks
* Sets up benchmark configs and directory structure to store weights (`operator/runtime/<RUNTIME>/weights`)
* (naive) fixes to some benchmarks: 
    * `pallet_datahaven_native_transfer`: 
      * use a mock for `NativeTokenId`
      * look at the balance difference of the treasury instead of the total (this makes the benchmark agnostic to genesis setup)
    * `snowbridge_pallet_system` / `snowbridge_pallet_system_v2` use native token xcm location vs relay chain one. Add missing benchmark methods and update fixture with valid data.
    * `snowbridge_pallet_ethereum_client`: update fixtures with valid data
    * `snowbrige_pallet_inbound_queue_v2`: set EthereumGatewayAddress when initializing storage on benchmark and use a mock message processor ( as fixture has `CreateAsset` payload which is not supported in the `EigenLayerMessageProcessor`)
    * `snowbridge_pallet_outbound_queue_v2`: add missing `submit_delivery_receipt` benchmark which required a dedicated fixture (all copied from the upstream  pallet)
    * `pallet_treasury`: Use an `ExistentialDeposit` of  `1` on benchmark, else payout fails.
    * `pallet_transaction_payment`: Use a custom `WeightToFee` that makes the Fee small, else account in benchmark cannot pay for fees (It is funded a multiplier of  `ExistentialDeposit` and is expected for that to be enough, but it's not in our particular setup).
 * comment out `pallet_identity` and `pallet_im_online` due to incompatibilities (to be addressed later)
 * Basic benchmark run to set `WeightInfo`  from `weights` in configs (real run should be done later using target hardware)


operator/scripts/run-benchmarks.sh
```diff
@@ -0,0 +1,167 @@
...
+fi
+
+echo -e "${GREEN}Found ${#PALLETS[@]} pallets to benchmark:${NC}"
+for pallet in "${PALLETS[@]}"; do
+    echo "  - $pallet"
+done
+echo ""
+
+# Track success/failure
+declare -A RESULTS
```
stiiifff, line 91: `-A` fails on MacOS ... I put a lowercase `-a` and then all is fine.

operator/README.md
```diff
@@ -2,6 +2,41 @@
 
 Based on [polkadot-sdk-solochain-template](https://github.com/paritytech/polkadot-sdk-solochain-template)
 
+## Benchmarking
+
+DataHaven uses runtime benchmarking to generate accurate weight calculations for all pallets. The benchmarking process is automated using `frame-omni-bencher`.
+
+### Requirements
+
+- `frame-omni-bencher` - Install with: `cargo install frame-omni-bencher --profile=production`
```
stiiifff: This didn't work .. I got an error that the `production` profile did not exist.

operator/runtime/mainnet/src/benchmarks.rs
```diff
@@ -24,16 +24,45 @@
...
+    [pallet_beefy_mmr, BeefyMmrLeaf]
+
+    // Substrate pallets
     [pallet_balances, Balances]
-    [pallet_identity, Identity]
-    [pallet_im_online, ImOnline]
+    // TODO: We cannot add due to EthereumSignature not being able
+    // to sign with Sr25519. Can be implemented once we incorporate
+    // https://github.com/paritytech/polkadot-sdk/pull/8179
+    // [pallet_identity, Identity]
```
undercover-cactus, line 37: Should we create a ticket for this TODO ?

operator/pallets/system/src/benchmarking.rs
```diff
@@ -74,8 +74,11 @@ mod benchmarks {
 
         T::Token::mint_into(&caller, amount)?;
 
-        let relay_token_asset_id: Location = Location::parent();
-        let asset = Box::new(VersionedLocation::from(relay_token_asset_id));
+        // TODO: Upstream pallet uses `Location::parent()` because it is a parachain
+        // working with the relay chain native token. We will need to adapt to that
+        // benchmark when we move to using the upstream pallet
+        let native_token_location: Location = Location::here();
+        let asset = Box::new(VersionedLocation::from(native_token_location));
```
undercover-cactus, line 81: Do we need to open a ticket for that ?

operator/runtime/mainnet/src/benchmarks.rs
```diff
@@ -24,16 +24,45 @@
...
+    // TODO: Benchmark does not work when MaxKeys is set < 1000.
+    // See https://github.com/paritytech/polkadot-sdk/blob/180fcce69fbddfa1bde9830362d8026340b4b750/substrate/frame/im-online/src/benchmarking.rs#L32
+    // [pallet_im_online, ImOnline]
```
undercover-cactus: How can we fix this ? What is the TODO here ?

operator/README.md
undercover-cactus: Same for me

operator/runtime/mainnet/src/benchmarks.rs
elfedy: @TDemeco submitted a fix upstream: https://github.com/paritytech/polkadot-sdk/pull/9325. We can add the pallet back once those changes are incorporated on the workspace's deps.

operator/README.md
elfedy: You folks can try `cargo install --git https://github.com/paritytech/polkadot-sdk frame-omni-bencher --profile=production --locked` else update the README to get the binary directly.

operator/runtime/mainnet/src/benchmarks.rs
undercover-cactus: Looks like they releases the fix in 2412 (https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2412-8).

I will reactivate this.

operator/runtime/mainnet/src/benchmarks.rs
gonzamontiel: I think we can already address this since Tobi backported a similar change to 2412: https://github.com/paritytech/polkadot-sdk/pull/9099

undercover-cactus: I tried and it worked for `IAmOline` but not for `Identity`

operator/runtime/testnet/src/configs/mod.rs
```diff
@@ -498,6 +507,18 @@ impl pallet_identity::Config for Runtime {
     type WeightInfo = ();
     type UsernameDeposit = ();
     type UsernameGracePeriod = ();
+
+    #[cfg(feature = "runtime-benchmarks")]
+    fn benchmark_helper(message: &[u8]) -> (Vec<u8>, Vec<u8>) {
+        let public = sp_io::crypto::ecdsa_generate(0.into(), None);
+        let eth_signer: Self::SigningPublicKey = public.into();
+        let hash_msg = sp_io::hashing::keccak_256(message);
+        let signature = Self::EthereumSignature::new(
```
TDemeco:
```suggestion
        let signature = Self::OffchainSignature::new(
```

PR #111, opened 2025-06-27
fix: :bug: set initial reward registry address to avoid solochain relayer error

TDemeco:
This PR solves an issue where the solochain relayer was fatally crashing because the DataHaven chain initialized with an empty `RewardsRegistryAddress` parameter. This caused all initial reward merkle root update messages to target the zero address before the parameter could be properly set by the launch command of the CLI. When the relayer tried to call the update function on the zero address (which isn't a valid contract), the transactions reverted and crashed the relayer.

The band-aid fix implemented is to set the actual `RewardsRegistry` contract address as the default value for the `RewardsRegistryAddress` parameter (since we know it as it's consistent between our CLI runs). This ensures initial messages have a valid target contract address from startup, preventing the fatal crashes.
Again, this is a temporary fix until we implement a more robust solution for parameter initialization, since it won't hold up if any changes happen to our contracts.

### DIFF
--- a/operator/runtime/stagenet/src/configs/runtime_params.rs
+++ b/operator/runtime/stagenet/src/configs/runtime_params.rs
@@ -1,10 +1,14 @@
-use crate::Runtime;
+use crate::{Balance, BlockNumber, Runtime, NANO_UNIT, UNIT};
 use frame_support::dynamic_params::{dynamic_pallet_params, dynamic_params};
 use hex_literal::hex;
 use sp_core::{ConstU32, H160, H256};
 use sp_runtime::{BoundedVec, Perbill};
 use sp_std::vec;
 
+#[cfg(feature = "storage-hub")]
+use crate::configs::storagehub::{ChallengeTicksTolerance, ReplicationTargetType, SpMinDeposit};
+
+#[cfg(not(feature = "storage-hub"))]
 #[dynamic_params(RuntimeParameters, pallet_parameters::Parameters::<Runtime>)]
 pub mod dynamic_params {
     use super::*;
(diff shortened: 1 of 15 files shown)


operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -19,14 +19,16 @@ pub mod dynamic_params {
         /// Set the initial address of the Snowbridge Gateway contract on Ethereum.
         /// The fact that this is a parameter means that we can set it initially to the zero address,
         /// and then change it later via governance, to the actual address of the deployed contract.
-        pub static EthereumGatewayAddress: H160 = H160::repeat_byte(0x0);
+        pub static EthereumGatewayAddress: H160 =
```
gonzamontiel: Let's revert this change since the problem is while running our CLI (only stagenet)

operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -19,14 +19,16 @@ pub mod dynamic_params {
...
+            H160::from_slice(&hex!("8f86403a4de0bb5791fa46b8e795c547942fe4cf"));
 
         #[codec(index = 1)]
         #[allow(non_upper_case_globals)]
         /// Set the initial address of the Rewards Registry contract on Ethereum.
         /// The fact that this is a parameter means that we can set it initially to the zero address,
         /// and then change it later via governance, to the actual address of the deployed contract.
-        pub static RewardsRegistryAddress: H160 = H160::repeat_byte(0x0);
```
gonzamontiel: Same here

operator/runtime/testnet/src/configs/runtime_params.rs
```diff
@@ -20,14 +20,16 @@ pub mod dynamic_params {
         /// Set the initial address of the Snowbridge Gateway contract on Ethereum.
         /// The fact that this is a parameter means that we can set it initially to the zero address,
         /// and then change it later via governance, to the actual address of the deployed contract.
-        pub static EthereumGatewayAddress: H160 = H160::repeat_byte(0x0);
```
gonzamontiel: Also let's revert it for testnet, as it may be a different address or we even wouldn't need to be automated like this.

.gitmodules
```diff
@@ -7,3 +7,4 @@
 [submodule "contracts/lib/snowbridge"]
 	path = contracts/lib/snowbridge
 	url = https://github.com/Moonsong-Labs/snowbridge
+	branch = solochain
```
ahmadkaouk: I have been actually working on bumping the snowbridge module in https://github.com/Moonsong-Labs/datahaven/pull/109. I was going to use the branch method, but after digging a bit, I saw that the recommended way is pinning to a commit as it make it easier to have deterministic, repeatable build espacially when start using multiple branches.

operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -36,10 +40,10 @@ pub mod dynamic_params {
 
         #[codec(index = 3)]
         #[allow(non_upper_case_globals)]
-        /// The RewardsAgentOrigin is the hash of the string "external_validators_rewards"
+        /// The RewardsAgentOrigin is the origin of the rewards agent, which is its ID.
         /// TODO: Decide which agent origin we want to use. Currently for testing it's the zero hash
         pub static RewardsAgentOrigin: H256 = H256::from_slice(&hex!(
-            "c505dfb2df107d106d08bd0f1a0acd10052ca9aa078629a4ccfd0c90c6e69b65"
+            "0000000000000000000000000000000000000000000000000000000000000000"
```
ahmadkaouk, line 50: Why this has been set to 00000

.gitmodules
TDemeco: Ok! I'll remove this change from here and you can add it to #109

operator/runtime/stagenet/src/configs/runtime_params.rs
TDemeco: Currently the agent ID of the rewards agent is the zero hash. This will probably be changed in the future but we haven't really decided a strategy for agent IDs yet

PR #113, opened 2025-07-01
Fix: 🏗️  Message encoding / decoding

gonzamontiel:
## Summary of changes
- We decided to remove the topics and nonce from the message encoding since we don't use them (original commit: https://github.com/Moonsong-Labs/datahaven/commit/ee2a3f2fd432e0f8623a9798730ce697bfc52d34).
  - Besides, we already have a nonce at the Snowbridge message level https://github.com/Moonsong-Labs/datahaven/blob/f4ab5c2b2e93c775555ab4b8d0b2babaf5bc15b7/operator/primitives/snowbridge/inbound-queue/src/v2/message.rs#L105

- I had to recreate the static test for _encoding_ (happens in [DataHavenSnowbridgeMessages.sol](https://github.com/Moonsong-Labs/datahaven/blob/d12d40634f16eb9642b5b94cbc364b6679339393/contracts/src/libraries/DataHavenSnowbridgeMessages.sol) ) / _decoding_ (happens in [operator/primitives/bridge/src/lib.rs)](https://github.com/Moonsong-Labs/datahaven/blob/f9f9cc65fe33c86ac91f2097adbcb945b1746277/operator/primitives/bridge/src/lib.rs). Now it matches the current structure. The idea is that now we can test that we don't break the decoding in followup refactoring.
- Fixes a problem with EigenLayer validator addresses. In all our contracts we were using `bytes32` to refer to a Solochain validator address. But on our Substrate change we actually expect AccountId20, so only 20 bytes. This was causing the decoding to fail. 
  - I opted for the minimal change that would be to take the right-most 20 bytes to send that to our chain. But we might want aswell to limit our EigenLayer contracts to be only 20 bytes long. @ahmadkaouk showcase this [here](https://github.com/Moonsong-Labs/datahaven/commit/92a34c273ca439341bd8b61ff3592cef45ba4727)
- Adds a bash script to run the static test. The test will compile the contracts, run the encoding test, compile the operator, and run the decoding test. This saves a huge amount of time since we don't need to run the full e2e setup. The way of running it is the following:
```bash
cd operator/test/scripts
./test_message_encoding.sh
```
- As a consequence of this PR, the execution relayer now works properly.

EDIT:

> [!IMPORTANT]
**We decided to use 20-byte addresses in our contracts**. So what is stated above is not valid anymore. 

The change implies that the mapping from Ethereum addresses to bytes32 addresses now it's a mapping as follows:

https://github.com/Moonsong-Labs/datahaven/blob/dd3ba99ac0553a33448dc77efbbea08cacd278e6/contracts/src/DataHavenServiceManager.sol#L51-L52

I've updated helper functions, tests, etc to be compliant with this change. The execution relayer and beefy relayer look stable now.

### DIFF
--- a/.github/workflows/actions/setup-env/action.yml
+++ b/.github/workflows/actions/setup-env/action.yml
@@ -25,21 +25,15 @@ runs:
     - name: Show sccache stats
       shell: bash
       run: sccache --show-stats
-    
-    - name: Cache cargo + target + sccache
-      uses: Swatinem/rust-cache@v2
+      
+    - uses: actions/cache@v4
       with:
-        cache-targets: true
-        cache-all-crates: true
-        shared-key: ${{ runner.os }}-${{inputs.cache-key}}-${{ hashFiles('operator/rust-toolchain.toml') }}-${{ hashFiles('operator/Cargo.lock') }}
-        workspaces: |
-          operator -> target
-        cache-on-failure: true
-        cache-bin: true
-        cache-directories: |
-          **/sccache         
-          **/.cargo/registry
-          **/.cargo/git
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
     
     - name: Setup Rust toolchain
       uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/task-build-operator.yml
+++ b/.github/workflows/task-build-operator.yml
@@ -1,6 +1,6 @@
 # Build Operator: CI for building the operator binary
 
-name: DataHave Operator Rust Tests
+name: DataHaven Operator Binary Build
 
 on:
   workflow_dispatch:


contracts/src/libraries/DataHavenSnowbridgeMessages.sol
```diff
@@ -20,16 +20,9 @@ library DataHavenSnowbridgeMessages {
     /**
      * @title New Validator Set Snowbridge Message
      * @notice A struct representing a new validator set to be sent as a message through Snowbridge.
-     *         This mimics the message format defined in the Snowbridge inbound pallet of the DataHaven
-     *         solochain.
-     * !IMPORTANT: The fields in this struct are placeholder until we have the actual message format
-     * !           defined in the DataHaven solochain.
+     *         This contains only the payload as nonce and topic are not used.
```
ahmadkaouk:
```suggestion
```

contracts/src/libraries/DataHavenSnowbridgeMessages.sol
```diff
@@ -45,6 +38,22 @@ library DataHavenSnowbridgeMessages {
...
+        bytes32[] memory validatorSet
+    ) internal pure returns (bytes20[] memory) {
+        bytes20[] memory validatorSetBytes20 = new bytes20[](validatorSet.length);
+        for (uint32 i = 0; i < validatorSet.length; i++) {
+            // Convert bytes32 to address (which is 20 bytes) by taking the rightmost 20 bytes
+            validatorSetBytes20[i] = bytes20(uint160(uint256(validatorSet[i])));
+        }
+        return validatorSetBytes20;
+    }
+
```
ahmadkaouk: The PR description states that the conversion uses the most significant 20 bytes, but in this implementation, you’re using the least significant 20 bytes. Can you clarify which approach is correct? Also, what was the reasoning behind choosing the least or most significant bytes for the conversion?

gonzamontiel: Sorry, my bad, I mean right-most bytes (fixed it in the desc). Since both solutions are equivalent, the main reason was for testing, if we use values such as `0x00...1` , then the conversion from 32bytes to 20 bytes is straight forward (taking always the right most significant bytes).

ahmadkaouk: We should update the implementation to use 20-byte addresses directly instead of relying on a conversion. While the current change may fix the encoding/decoding issue, it would still result in invalid validator addresses being included in the ValidatorSet message.

gonzamontiel: Implemented.

contracts/config/example.jsonc
```diff
@@ -98,13 +98,13 @@
     /// The origin linked to the Rewards Agent, the Agent contract who's allowed to submit
     /// new reward merkle roots to the RewardsRegistry contract.
     "rewardsMessageOrigin": "0x0000000000000000000000000000000000000000000000000000000000000000",
-    /// The initial validators for the BEEFY light client.
-    "initialValidators": [
+    /// The initial validator hashes for the BEEFY light client.
```
ahmadkaouk, line 101: It’s a personal preference, but I find using addresses instead of hashes makes it easier to reason about and debug.

PR #114, opened 2025-07-09
ci: caching between run in the CI to make building faster

undercover-cactus:
~~Improve CI runs by adding a caching action. Between each run crates should be cached to speed up building datahaven binary.~~

## Summary

In the end the real issue was not that we were missing a caching action but that we were caching too much. We went way over the 10GB limit imposed by github (we were using 60GB of cache). So the most recent cache were deleted right away or not cache at all.

The over use of the cache was happening because we were caching twice sccache folder. Once with the `mozilla-actions/sccache-action` and an other time with `Swatinem/rust-cache`.

The solution was to keep caching sccache on github with `mozilla-actions/sccache-action`.

In this PR we also exchange the `Swatinem/rust-cache` action to use the more standard `actions/cache@v4`. It would avoid in the future unwanted breaking changes and security issues.

### DIFF
--- a/contracts/src/interfaces/IRewardsRegistry.sol
+++ b/contracts/src/interfaces/IRewardsRegistry.sol
@@ -9,14 +9,18 @@ interface IRewardsRegistryErrors {
     error OnlyAVS();
     /// @notice Thrown when a function is called by an address that is not the RewardsAgent.
     error OnlyRewardsAgent();
-    /// @notice Thrown when rewards have already been claimed for the current merkle root.
-    error RewardsAlreadyClaimed();
     /// @notice Thrown when a provided merkle proof is invalid.
     error InvalidMerkleProof();
     /// @notice Thrown when rewards transfer fails.
     error RewardsTransferFailed();
     /// @notice Thrown when the rewards merkle root is not set.
     error RewardsMerkleRootNotSet();
+    /// @notice Thrown when trying to access a merkle root index that doesn't exist.
+    error InvalidMerkleRootIndex();
+    /// @notice Thrown when trying to claim rewards for a root index that has already been claimed.
+    error RewardsAlreadyClaimedForIndex();
+    /// @notice Thrown when the arrays provided to the batch claim function have mismatched lengths.
+    error ArrayLengthMismatch();
 }
 
 /**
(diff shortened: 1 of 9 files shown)


.github/workflows/task-build-operator.yml
```diff
@@ -38,7 +38,15 @@ jobs:
...
-
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
ahmadkaouk:
```suggestion
```
We already have comprehensive caching configured :
https://github.com/Moonsong-Labs/datahaven/blob/42dd98e38c8f84eee432de8474f3719357972307/.github/workflows/actions/setup-env/action.yml#L29-L42
Adding actions/cache@v4 would duplicate these caches.

.github/workflows/task-build-operator.yml
```diff
@@ -1,6 +1,6 @@
 # Build Operator: CI for building the operator binary
 
-name: DataHave Operator Rust Tests
+name: DataHaven Operator Rust Tests
```
ahmadkaouk:
```suggestion
name: DataHaven Operator Binary Build
```
The workflow is building the binary not running the Rust tests

.github/workflows/task-build-operator.yml
undercover-cactus: Yeah it looks like it doesn't work.

https://github.com/Moonsong-Labs/datahaven/actions/runs/16188125641/job/45697601700#step:3:177

.github/workflows/actions/setup-env/action.yml
```diff
@@ -25,21 +25,16 @@ runs:
...
-        shared-key: ${{ runner.os }}-${{inputs.cache-key}}-${{ hashFiles('operator/rust-toolchain.toml') }}-${{ hashFiles('operator/Cargo.lock') }}
-        workspaces: |
-          operator -> target
-        cache-on-failure: true
-        cache-bin: true
-        cache-directories: |
-          **/sccache         
-          **/.cargo/registry
-          **/.cargo/git
+        path: |
```
timbrinded, line 31: `cache-on-failure` can be quite useful, since a failing test in a big refactor PR means you wont want to wait for it to rebuild from scratch each time.

.github/workflows/actions/setup-env/action.yml
```diff
@@ -25,21 +25,16 @@ runs:
...
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
gonzamontiel, line 36: Why removing `${{ hashFiles('operator/rust-toolchain.toml') }}` ? I think we can leave the key as it was.

.github/workflows/actions/setup-env/action.yml
gonzamontiel: I liked the cargo oriented cache,  but there's no way of proving that this simple approach is better other than test it for some time.

.github/workflows/actions/setup-env/action.yml
undercover-cactus: It is mostly relevant for sccache if I am not mistaken and it is not the responsibility of this action anymore.

.github/workflows/actions/setup-env/action.yml
undercover-cactus: It is a wrapper around github cache action. It might bring some shortcut but it also obfuscates behaviors.
Also the main reason for removing it is for security concerns.

@timbrinded Indeed it could be frustrating that if after changing dependencies the build fails and we have to restart from scratch.
I have found this case describe in the documentation here : https://github.com/actions/cache/blob/main/save/README.md#always-save-cache

I will try it.

PR #120, opened 2025-07-22
test: Native token transfer e2e tests

ahmadkaouk:
### Summary
- **Add** `test/suites/native-token-transfer.test.ts` focused on the HAVE native token lifecycle via Snowbridge v2.
- **Validate** registration, DataHaven → Ethereum mints, Ethereum → DataHaven unlocks, event emission, and 1:1 backing invariant.

### Tests added
- should register DataHaven native token on Ethereum
- should transfer tokens from DataHaven to Ethereum
- should maintain 1:1 backing ratio
- should emit transfer events
- should transfer tokens from Ethereum to DataHaven (Snowbridge v2)

### What the suite covers
- **Registration**: Sudo-registers the native token; confirms `ForeignTokenRegistered` on the Gateway; verifies ERC-20 metadata (`HAVE`/`wHAVE`, 18 decimals).
- **DataHaven → Ethereum**: Executes `transfer_to_ethereum`; asserts Substrate events (`TokensLocked`, `TokensTransferredToEthereum`); observes Ethereum `Transfer` mint (from zero address); validates sender balance delta, sovereign account increase, and ERC-20 recipient credit.
- **Backing invariant**: Ensures sovereign account balance ≥ ERC-20 total supply.
- **Event emission**: Confirms key Substrate events without polling delays.
- **Ethereum → DataHaven**: Approves and calls `Gateway.sendToken`; if unsupported locally, the test skips; otherwise asserts burn on Ethereum and unlock on DataHaven with corresponding balance deltas.


CLAUDE.md
undercover-cactus, line 1: Do we really need this file in the repo ? It is not related to DataHaven.

test/suites/native-token-transfer.test.ts
```diff
@@ -0,0 +1,605 @@
...
+    const tokenAddress = (await connectors.publicClient.readContract({
+      address: deployments.Gateway,
+      abi: gatewayAbi,
+      functionName: "tokenAddressOf",
+      args: [tokenId]
+    })) as `0x${string}`;
+
+    return tokenAddress === ZERO_ADDRESS ? null : tokenAddress;
+  } catch (_error) {
+    return null;
```
undercover-cactus: On line 97, we are throwing an error but then catching it here and not doing anything with the message at all. Should we print the error for better understanding of what is happening ?

We could just make it fails too. We are in a test failing loudly is fine.

test/suites/native-token-transfer.test.ts
```diff
@@ -0,0 +1,605 @@
...
+    deployments = await parseDeploymentsFile();
+  });
+  it("should register DataHaven native token on Ethereum", async () => {
+    const connectors = suite.getTestConnectors();
+    // First, check if token is already registered
+    const existingTokenAddress = await getNativeERC20Address(connectors);
+
+    // Skip registration if token already exists
+    if (existingTokenAddress) {
+      logger.debug(`Token already registered at: ${existingTokenAddress}`);
```
undercover-cactus: In which scenario we would have not the token registered and in an other run have it registered ?

CLAUDE.md
ahmadkaouk: This is particularly useful when using Claude Code. It provides more context about the repo and yields better results. I added it because many of us are using Claude Code, so it might also be beneficial for them.

test/suites/native-token-transfer.test.ts
ahmadkaouk: Initially, I made this change to speed up local debugging and allow reusing the same infra between re-runs. That said, for this suite we want to validate registration from a clean state. I’ll remove the ‘already registered’ guard and make the test fail if the token is pre-registered, since it shouldn’t be at the beginning.

test/suites/native-token-transfer.test.ts
ahmadkaouk: I agree the try/catch is not needed here. Changed in 8f214f3860ab4192598b7e9769d3cf435433f61e

PR #126, opened 2025-07-31
test: Update validator set e2e test

gonzamontiel:
## Add E2E validator-set update flow

- feat: `test/utils/validators.ts` for on-demand validator orchestration.
- feat: `test/suites/validator-set-update.test.ts` covering allowlist → register → update.
- some minor launcher updates: avoid docker cache, add `--platform` when building datahaven image, avoid sending validator-set update on launch.
- Helpers: ABI shortcut in `test/utils/contracts.ts`; config tweaks in `test/configs/validator-set.json`.
- Minor cleanup/formatting across `test/launcher/*`, `test/scripts/setup-validators.ts`, and related tests.
- added `keepAlive` flag to `BaseTestSuite`, in order to avoid tearing down the network while debugging. Defaults, obviously, to false.
- added a `failOnTomeout` option on to waitForDataHavenEvents() so the test fails of the timeout is reached and no event was captured.

### Coverage
- The test simulates an scenario in which we have two active authorities (alice and bob), which are running, and registered as operators, which is the normal state after the chain launches. Then:
- It launches two more nodes (charlie and dave)
- It add the nodes to allowlist and register them as operators
- It sends the validator set update message
- Checks that the validator update message was propagated through the gateway and arrived the external-validators pallet
- Checks that the chain continues producing blocks
 
### Notes
The last test case has a timeout of 10 minutes. This is to respect propagation times of the message through the relayers. We are testing that the external validators pallet actually updated the validator set. Locally, I could expect 5~6 minutes, I just wanted to be on the safe side. CI is passing showing that this was enough indeed.

### DIFF
--- a/operator/node/src/command.rs
+++ b/operator/node/src/command.rs
@@ -113,7 +113,8 @@ macro_rules! construct_async_run {
 				runner.async_run(|$config| {
 					let $components = service::new_partial::<datahaven_mainnet_runtime::Runtime, datahaven_mainnet_runtime::RuntimeApi>(
 						&$config,
-                        &mut $cli.eth.clone()
+                        &mut $cli.eth.clone(),
+                        false,
 					)?;
 					let task_manager = $components.task_manager;
 					{ $( $code )* }.map(|v| (v, task_manager))
@@ -123,7 +124,8 @@ macro_rules! construct_async_run {
                 runner.async_run(|$config| {
 					let $components = service::new_partial::<datahaven_testnet_runtime::Runtime, datahaven_testnet_runtime::RuntimeApi>(
 						&$config,
-                        &mut $cli.eth.clone()
+                        &mut $cli.eth.clone(),
+                        false,
 					)?;
 					let task_manager = $components.task_manager;
 					{ $( $code )* }.map(|v| (v, task_manager))
@@ -133,7 +135,8 @@ macro_rules! construct_async_run {
 				runner.async_run(|$config| {
 					let $components = service::new_partial::<datahaven_stagenet_runtime::Runtime, datahaven_stagenet_runtime::RuntimeApi>(
 						&$config,
-                        &mut $cli.eth.clone()
+                        &mut $cli.eth.clone(),
+                        false,
 					)?;
 					let task_manager = $components.task_manager;
 					{ $( $code )* }.map(|v| (v, task_manager))
@@ -150,21 +153,21 @@ macro_rules! construct_benchmark_partials {
                 let $partials = service::new_partial::<
                     datahaven_mainnet_runtime::Runtime,
                     datahaven_mainnet_runtime::RuntimeApi,
-                >(&$config, &mut $cli.eth.clone())?;
+                >(&$config, &mut $cli.eth.clone(), false)?;
                 $code
             }
             ref spec if spec.is_testnet() => {
                 let $partials = service::new_partial::<
                     datahaven_testnet_runtime::Runtime,
                     datahaven_testnet_runtime::RuntimeApi,
-                >(&$config, &mut $cli.eth.clone())?;
+                >(&$config, &mut $cli.eth.clone(), false)?;
                 $code
             }
             _ => {
                 let $partials = service::new_partial::<
                     datahaven_stagenet_runtime::Runtime,
                     datahaven_stagenet_runtime::RuntimeApi,
-                >(&$config, &mut $cli.eth.clone())?;
+                >(&$config, &mut $cli.eth.clone(), false)?;
                 $code
             }
         }
--- a/operator/node/src/service.rs
+++ b/operator/node/src/service.rs
@@ -243,28 +243,42 @@ where
 
 fn build_babe_inherent_providers(
     slot_duration: sp_consensus_babe::SlotDuration,
+    use_mock_timestamp: bool,
 ) -> (
     sp_consensus_babe::inherents::InherentDataProvider,
     sp_timestamp::InherentDataProvider,
 ) {
-    // In manual/instant sealing we want to advance time deterministically per block
-    // to satisfy `pallet_timestamp` MinimumPeriod without sleeping. We increment a
-    // static counter by one slot each time and use that value as the timestamp.
-    let increment = slot_duration.as_millis();
-    let next_ts = MOCK_TIMESTAMP
-        .fetch_add(increment, Ordering::SeqCst)
-        .saturating_add(increment);
-    let timestamp = sp_timestamp::InherentDataProvider::new(sp_timestamp::Timestamp::new(next_ts));
-    let slot = sp_consensus_babe::inherents::InherentDataProvider::from_timestamp_and_slot_duration(
-        *timestamp,
-        slot_duration,
-    );
-    (slot, timestamp)
+    if use_mock_timestamp {
+        // In manual/instant sealing we want to advance time deterministically per block
+        // to satisfy `pallet_timestamp` MinimumPeriod without sleeping. We increment a
+        // static counter by one slot each time and use that value as the timestamp.
+        let increment = slot_duration.as_millis();
+        let next_ts = MOCK_TIMESTAMP
+            .fetch_add(increment, Ordering::SeqCst)
+            .saturating_add(increment);
+        let timestamp =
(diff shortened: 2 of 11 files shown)


test/suites/validator-set-update.test.ts
```diff
@@ -0,0 +1,423 @@
...
+      event: "ExternalValidatorsSet",
+      timeout: 600_000,
+      failOnTimeout: true
+    });
+
+    if (!externalValidatorsSetEvent.data) {
+      logger.error("ExternalValidatorsSet event not found");
+      throw new Error("ExternalValidatorsSet event not found");
+    }
+    logger.success("ExternalValidatorsSet event found");
```
stiiifff, line 366: Should the logic also verify the expected (updated) validator set ?

gonzamontiel: Yes it was my intention, but I removed a part where the session actually has the 4 validators because I needed to wait another like 3/4 minutes to observe that. But once the event is received, the storage in the external validators pallet is updated. We can improve it later, though.

ahmadkaouk: You can add a check that query the `ExternalValidators` storage map in the ExternalValidators pallet, and verify if the validator set is correctly updated

test/suites/validator-set-update.test.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    // Add Charlie and Dave to the allowlist
+    for (let i = 0; i < newValidators.length; i++) {
+      const validator = newValidators[i];
+      logger.info(
+        `🔧 Adding validator ${i + 1}/${newValidators.length} (${validator.publicKey}) to allowlist...`
+      );
+
+      let retryCount = 0;
+      const maxRetries = 3;
```
ahmadkaouk: Why do we need a retry mechanism? Shouldn’t a transaction succeed on its first attempt? I don’t understand how a retry mechanism could resolve the problem. Also, the retry is not working. You are incrementing the `retryCount`, but there is no inner loop to retry.

test/suites/validator-set-update.test.ts
```diff
@@ -0,0 +1,423 @@
...
+        logger.info(`📝 Transaction hash for allowlist: ${hash}`);
+
+        const receipt = await connectors.publicClient.waitForTransactionReceipt({ hash });
+        logger.info(
+          `📋 Allowlist transaction receipt: status=${receipt.status}, gasUsed=${receipt.gasUsed}`
+        );
+
+        if (receipt.status === "success") {
+          logger.success(`Added ${validator.publicKey} to allowlist`);
+          break; // Success, exit retry loop
```
ahmadkaouk: The break here exits the outer validators loop, not a retry loop, so after the first successful validator the loop stops and the remaining validators won’t be processed

test/suites/validator-set-update.test.ts
```diff
@@ -0,0 +1,423 @@
...
+
+        if (retryCount >= maxRetries) {
+          logger.error(`All retry attempts failed for ${validator.publicKey}`);
+          throw error;
+        }
+
+        // Wait before retry
+        await new Promise((resolve) => setTimeout(resolve, 2000 * retryCount));
+      }
+    }
```
ahmadkaouk: After allowlisting, there is no check that both Charlie and Dave were successfully allowlisted. You could read validatorsAllowlist(address).

test/suites/validator-set-update.test.ts
```diff
@@ -0,0 +1,423 @@
...
+
+        const receipt = await connectors.publicClient.waitForTransactionReceipt({ hash });
+        logger.info(
+          `📋 Operator set registration receipt: status=${receipt.status}, gasUsed=${receipt.gasUsed}`
+        );
+
+        if (receipt.status === "success") {
+          logger.success(`Registered ${validator.publicKey} for operator sets`);
+          break; // Success, exit retry loop
+        }
```
ahmadkaouk: We can add a check that validatorEthAddressToSolochainAddress was updated for the new validators

test/cli/handlers/launch/index.ts
```diff
@@ -95,8 +95,6 @@ const launchFunction = async (options: LaunchOptions, launchedNetwork: LaunchedN
 
   await launchRelayers(options, launchedNetwork);
 
-  await performValidatorSetUpdate(options, launchedNetwork.elRpcUrl, contractsDeployed);
-
```
ahmadkaouk, line 99: Why removing this ? it will break the `--update-vaidator-set` flag that do nothing now

test/suites/validator-set-update.test.ts
gonzamontiel: This was obsolete while I was debugging an issue with the relayers, I removed the inner loop but still some stuff was not deleted, I'll remove all the left overs.

test/cli/handlers/launch/index.ts
undercover-cactus: Is it because now we want to have the liberty to do it in the e2e tests ?

test/framework/suite.ts
```diff
@@ -70,6 +81,11 @@ export abstract class BaseTestSuite {
       logger.info(`🧹 Tearing down test suite: ${this.options.suiteName}`);
 
       try {
+        if (this.options.keepAlive && !isCI) {
+          this.printNetworkInfo();
+          await this.waitForEnter();
```
undercover-cactus, line 86: I am not convince by this but I agree that we would need something to help with debugging. The docker containers are removed right after the tests and you can't read the logs in case something went wrong.

Is it the reason why you added this feature ?

test/launcher/network/index.ts
```diff
@@ -172,7 +172,7 @@ export const launchNetwork = async (
         datahavenImageTag: options.datahavenImageTag || "datahavenxyz/datahaven:local",
         relayerImageTag: options.relayerImageTag || "datahavenxyz/snowbridge-relay:latest",
         authorityIds: TEST_AUTHORITY_IDS,
-        buildDatahaven: options.buildDatahaven ?? true,
+        buildDatahaven: options.buildDatahaven ?? !isCI, // if not specified, default to false for CI, true for local testing
```
undercover-cactus, line 175: This is indeed needed but it makes the testing framework more confusing with different behaviors for different environments. In the future, we should remove this.

test/suites/validator-set-update.test.ts
gonzamontiel: Same as above. will fix it  : )

test/framework/suite.ts
gonzamontiel: Yup! I explained in the description

> added keepAlive flag to BaseTestSuite, in order to avoid tearing down the network while debugging. Defaults, obviously, to false.

test/launcher/network/index.ts
gonzamontiel: I think is not really much less clear, for readability, we could have a Default() and then fill this default depending on the env variables. The result ill be the same but maybe more readable.

test/cli/handlers/launch/index.ts
gonzamontiel: Yes, it was really unnecessary to perform that change on a normal launch, since the current validators now match the current operators in the contracts. And yes, we are testing explicitly on the e2e tests.

gonzamontiel: > Why removing this ? it will break the --update-vaidator-set flag that do nothing now

I'll remove the flag, we don't really need that if it isn't for testing

test/package.json
```diff
@@ -7,7 +7,7 @@
     "cli": "bun run cli/index.ts",
     "fmt": "biome check .",
     "fmt:fix": "biome check --write .",
-    "build:docker:operator": "docker build -t datahavenxyz/datahaven:local -f ./docker/datahaven-node-local.dockerfile ../.",
+    "build:docker:operator": "docker build --no-cache --platform linux/amd64 -t datahavenxyz/datahaven:local -f ./docker/datahaven-node-local.dockerfile ../.",
```
ahmadkaouk, line 10: Is the `--no-cache` really necessary ? It will slow down the build

gonzamontiel: Uhm, yes, but this is used exclusively for building the local image while testing. What we usually want is that the image is not cached, so we have the binary of the code we're looking at, and not some cached image. It will slow down in cases where there are no changes, but for a few seconds.

PR #129, opened 2025-08-11
feat: ✨ Add custom chainspec support to DataHaven CLI

stiiifff:
## Summary

Adds `--chainspec` parameter to the DataHaven CLI deploy command for using custom chainspec files across all environments.

## Usage

```bash
# Deploy with custom chainspec
bun cli deploy --environment testnet --chainspec /absolute/path/to/chainspec.json

# Normal deployment (unchanged)
bun cli deploy --environment testnet
```

## Changes

- **CLI**: Added `--chainspec <value>` parameter with absolute path validation
- **Helm**: New ConfigMap template and init container for custom chainspecs
- **Bootnode**: Conditionally uses custom chainspec or generates dynamically
- **Distribution**: Bootnode serves chainspec via HTTP, validators download from bootnode


test/cli/handlers/deploy/datahaven.ts
```diff
@@ -62,30 +74,53 @@ export const deployDataHavenSolochain = async (
...
+
+  // Add custom chainspec configuration if provided
+  if (options.chainspec) {
+    logger.info(`🔗 Using custom chainspec: ${options.chainspec}`);
+    bootnodeArgs.push("--set-file", `customChainspecContent=${options.chainspec}`);
+  }
+
+  bootnodeArgs.push("-n", launchedNetwork.kubeNamespace, "--wait", "--timeout", bootnodeTimeout);
+
+  logger.debug(await $`helm ${bootnodeArgs}`.cwd(path.join(process.cwd(), "../deploy")).text());
```
undercover-cactus, line 98: Will this throw if there is an error ?

I have seen this somewhere else in the CLI. The command didn't throw and error, showing the next log line that announced that the deployment was successful.

undercover-cactus: I checked and it will throw. The only time it won't throw is when `nothrow()` is being called after.

PR #134, opened 2025-08-26
fix: Complete CI compatibility with self-hosted GitHub runners

stiiifff:
## Summary

This PR resolves all CI failures following the migration to self-hosted GitHub runners (`DH-Testing` group) by eliminating sudo dependencies and fixing Docker connectivity issues.

### 🔧 **Eliminated sudo requirements across all workflows**
- **Setup Environment**: Installed mold linker and system dependencies in userspace without sudo
- **Tool Installation**: Replaced apt/system package installations with direct binary downloads:
  - Kurtosis: Direct binary download from GitHub releases (v1.10.3)
  - Taplo: Direct binary installation for Cargo.toml formatting
  - cargo-nextest: Using `cargo install` instead of GitHub action (v0.9.100)
- **Runner Cleanup**: Skipped cleanup-runner action entirely on self-hosted runners (bare-metal manages disk space externally)

### 🐳 **Fixed Docker connectivity for E2E tests**  
- **Enhanced dockerode configuration** with robust fallback logic for different socket locations
- **Added DOCKER_HOST environment variable** to E2E workflow for consistent Docker daemon access
- **Implemented connection testing** with detailed error diagnostics for troubleshooting
- **Resolves FailedToOpenSocket errors** by supporting multiple socket paths and connection methods

### 🏷️ **Workflow optimizations**
- **Label-based targeting**: All heavy workloads (Rust builds, E2E tests) now run on `DH-Testing` runners
- **Dependency management**: Used `install-deps: false` flag instead of hardcoded runner detection  
- **Permission fixes**: Corrected Docker build permissions and GHCR organization names

### DIFF
--- a/.github/workflows/task-e2e.yml
+++ b/.github/workflows/task-e2e.yml
@@ -51,8 +51,8 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
       - name: Pull Kurtosis images
         run: |
-          docker pull ${{ env.KURTOSIS_CORE_IMAGE }}:{{ env.KURTOSIS_VERSION }}
-          docker pull ${{ env.KURTOSIS_ENGINE_IMAGE }}:{{ env.KURTOSIS_VERSION }}
+          docker pull ${{ env.KURTOSIS_CORE_IMAGE }}:${{ env.KURTOSIS_VERSION }}
+          docker pull ${{ env.KURTOSIS_ENGINE_IMAGE }}:${{ env.KURTOSIS_VERSION }}
       - name: Install Kurtosis
         run: |
           # Install Kurtosis locally without sudo


test/utils/docker.ts
```diff
@@ -5,42 +5,48 @@ import invariant from "tiny-invariant";
...
-    }
-    if (dockerHost.startsWith("tcp://")) {
-      const url = new URL(dockerHost);
-      return new Docker({
-        host: url.hostname,
-        port: Number.parseInt(url.port) || 2375,
-        protocol: "http"
-      });
-    }
+  // const dockerHost = process.env.DOCKER_HOST;
```
gonzamontiel: I think this was done as a test, but the CI is not passing, isn't it? We should revert this commit

undercover-cactus: Yes it is me. We have an issue with the CI not able to list containers. It seems to not be able to connect to the engine with the unix socket. I think the runner is using podman and not docker directly... I have another branch where I am testing other way to connect.

PR #138, opened 2025-08-28
feat: ✨ Add Moonbeam Batch precompile

stiiifff:
## Summary

This PR adds Moonbeam's Batch precompile to DataHaven, enabling efficient batching of multiple EVM transactions in a single call. This implementation follows Moonbeam's exact architecture and provides significant gas savings for batch operations.

### Batch Precompile Functions
- **`batchSome(address[], uint256[], bytes[], uint64[])`**: Execute multiple calls, continuing on failures
- **`batchSomeUntilFailure(address[], uint256[], bytes[], uint64[])`**: Execute calls until first failure
- **`batchAll(address[], uint256[], bytes[], uint64[])`**: Execute all calls, reverting if any fail

### Technical Implementation
- **Address**: `0x0808` (2056 in decimal)
- **Access Control**: Restricted nesting with `SubcallWithMaxNesting<2>`
- **Self-recursion**: Only the Batch precompile can call itself (`OnlyFrom<AddressU64<2056>>`)
- **Gas Management**: Proper gas estimation and refund handling
- **Error Handling**: Comprehensive revert reasons and event logging

Depends on #137

### DIFF
--- a/operator/Cargo.lock
+++ b/operator/Cargo.lock
@@ -1163,14 +1163,14 @@ source = "git+https://github.com/paritytech/polkadot-sdk?tag=polkadot-stable2412
 dependencies = [
  "bp-runtime",
  "finality-grandpa",
- "frame-support 39.1.0 (git+https://github.com/paritytech/polkadot-sdk?tag=polkadot-stable2412-6)",
+ "frame-support",
  "parity-scale-codec",
  "scale-info",
  "serde",
  "sp-consensus-grandpa",
- "sp-core 35.0.0 (git+https://github.com/paritytech/polkadot-sdk?tag=polkadot-stable2412-6)",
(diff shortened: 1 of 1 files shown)


operator/runtime/stagenet/src/precompiles.rs
```diff
@@ -44,6 +45,15 @@ type DataHavenPrecompilesAt<R> = (
     PrecompileAt<AddressU64<1026>, ECRecoverPublicKey, (CallableByContract, CallableByPrecompile)>,
     RemovedPrecompileAt<AddressU64<1027>>,
```
ahmadkaouk, line 46: Not related to changes in this PR but we should probably remove this line. (I will do it in my PR)

PR #148, opened 2025-09-04
fix: 🔧  implement babe randomness for BabeDataGetter

gonzamontiel:
Since we are including Storage Hub pallets and we no longer hide the compilation behind a feature flag, the tests started to fail because the nodes where panic!ing on each block finalisation.

### DIFF
--- a/operator/Cargo.lock
+++ b/operator/Cargo.lock
@@ -2386,6 +2386,7 @@ dependencies = [
  "pallet-ethereum",
  "pallet-evm",
  "pallet-evm-chain-id",
+ "pallet-evm-precompile-batch",
  "pallet-evm-precompile-blake2",
  "pallet-evm-precompile-bn128",
  "pallet-evm-precompile-modexp",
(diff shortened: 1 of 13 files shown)


operator/runtime/stagenet/src/configs/storagehub/mod.rs
```diff
@@ -109,13 +109,13 @@ impl pallet_randomness::Config for Runtime {
...
-        todo!("implement `get_epoch_index`");
+        pallet_babe::Pallet::<Runtime>::current_epoch().epoch_index
     }
     fn get_epoch_randomness() -> Hash {
-        todo!("implement `get_epoch_randomness`");
+        sp_core::H256::from(pallet_babe::Pallet::<Runtime>::current_epoch().randomness)
     }
     fn get_parent_randomness() -> Hash {
-        todo!("implement `get_parent_randomness`");
+        sp_core::H256::from(pallet_babe::Pallet::<Runtime>::current_epoch().randomness)
```
ahmadkaouk, line 118: Why the implementation of `get_parent_randomness()` is different from what's is implemented for mainnet ?
https://github.com/datahaven-xyz/datahaven/blob/1def9211b5dfebea7fdb32c2b25f86a7e9d8d8ab/operator/runtime/mainnet/src/configs/storagehub/mod.rs#L117-L121

stiiifff: @gonzamontiel Can you align all 3 runtimes on the same implementation ?

PR #149, opened 2025-09-05
feat: storage hub client

undercover-cactus:


operator/runtime/mainnet/Cargo.toml
```diff
@@ -266,6 +267,8 @@ std = [
 	"pallet-randomness/std",
 	"pallet-storage-providers/std",
 	"pallet-storage-providers-runtime-api/std",
+    "dep:shc-common",
+	"shc-common/std",
```
stiiifff, line 271: Why the double declaration here ?

operator/Cargo.toml
```diff
@@ -207,7 +212,7 @@ snowbridge-verification-primitives = { path = "primitives/snowbridge/verificatio
 
 # Frontier (wasm)
 evm = { git = "https://github.com/rust-ethereum/evm", rev = "6d86fe2d3bcc14887c2575f62958a67ac2d523db", default-features = false }
-fp-account = { git = "https://github.com/polkadot-evm/frontier", rev="75329a2df49e2cc7981485392c31160929d1bd48", default-features = false }
+fp-account = { git = "https://github.com/polkadot-evm/frontier", branch = "stable2412", default-features = false }
```
stiiifff, line 215: I don't understand why this ref needs to be changed.

operator/Cargo.toml
```diff
@@ -109,6 +111,7 @@ pallet-multisig = { git = "https://github.com/paritytech/polkadot-sdk", tag = "p
 pallet-offences = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
 pallet-parameters = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
 pallet-preimage = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
+pallet-collator-selection = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-stable2412-6", default-features = false }
```
stiiifff, line 114: This doesn't seem right, DataHaven doesn't need / use this.

operator/Cargo.toml
```diff
@@ -125,6 +128,8 @@ pallet-treasury = { git = "https://github.com/paritytech/polkadot-sdk", tag = "p
 pallet-utility = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
 pallet-whitelist = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
 pallet-xcm = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2412-6", default-features = false }
+parachain-info = { package = "staging-parachain-info", git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-stable2412-6", default-features = false }
+parachains-common = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-stable2412-6", default-features = false }
```
stiiifff, line 132: Why do we need to reference parachain stuff ?

operator/Cargo.lock
```diff
@@ -2060,95 +2222,388 @@ dependencies = [
...
  "sp-version",
- "thiserror 1.0.69",
+ "tracing",
 ]
 
 [[package]]
-name = "cumulus-test-relay-sproof-builder"
-version = "0.17.0"
-source = "git+https://github.com/paritytech/polkadot-sdk?tag=polkadot-stable2412-6#bbc435c7667d3283ba280a8fec44676357392753"
+name = "cumulus-client-parachain-inherent"
```
stiiifff, line 2321: I wonder why we have to pull Cumulus -related dependencies ..

operator/runtime/mainnet/Cargo.toml
undercover-cactus: If I am not mistaken it is because the `shc-common` is optional and by having `dep:shc-common` we are requiring it to be present. And `shc-common/std` is activating the `std` feature for this library. 

If you look at line 140, you can see the dependency
```
shc-common = { workspace = true, optional = true }
```

operator/Cargo.toml
undercover-cactus: Yes, it is a dependency from SH that wouldn't compile. It is residual from tests and should be remove.

operator/Cargo.toml
undercover-cactus: Same as `pallet-collator-selection`.

operator/Cargo.lock
undercover-cactus: We do need this to have the same client as SH otherwise we can't use SH client. This is really important.

operator/node/src/command.rs
```diff
@@ -270,23 +327,26 @@ pub fn run() -> sc_cli::Result<()> {
...
                                     sc_network::NetworkWorker<_, _>,
-                                >(config, cli.eth)
+                                >(config, cli.eth, provider_options)
                                 .await
                             }
                             ref spec if spec.is_testnet() => {
                                 service::new_full::<
-                                    datahaven_testnet_runtime::RuntimeApi,
+                                    datahaven_mainnet_runtime::Runtime,
+                                    datahaven_mainnet_runtime::RuntimeApi,
```
ahmadkaouk, line 339:
```suggestion
                                    datahaven_testnet_runtime::Runtime,
                                    datahaven_testnet_runtime::RuntimeApi,
```

operator/node/src/command.rs
```diff
@@ -270,23 +327,26 @@ pub fn run() -> sc_cli::Result<()> {
...
                                     sc_network::NetworkWorker<_, _>,
-                                >(config, cli.eth)
+                                >(config, cli.eth, provider_options)
                                 .await
                             }
                             _ => {
                                 service::new_full::<
-                                    datahaven_stagenet_runtime::RuntimeApi,
+                                    datahaven_mainnet_runtime::Runtime,
+                                    datahaven_mainnet_runtime::RuntimeApi,
```
ahmadkaouk, line 347:
```suggestion
                                    datahaven_stagenet_runtime::Runtime,
                                    datahaven_stagenet_runtime::RuntimeApi,
```

ahmadkaouk: PR to fix it https://github.com/datahaven-xyz/datahaven/pull/157

PR #153, opened 2025-09-08
feat: ✨ Optimize E2E tests with shared network instance

stiiifff:
## Summary
This PR optimizes the E2E test execution by implementing a shared network instance across all test suites, preventing Kurtosis from being restarted for each suite.

## Problem
Previously, each test suite would launch its own complete network stack including:
- Kurtosis Ethereum network
- DataHaven nodes
- Smart contracts deployment
- Snowbridge relayers

This resulted in significant overhead and slow test execution times as the entire infrastructure was torn down and recreated between test suites.

## Solution
Implemented a shared network architecture that:
- Launches the network infrastructure once at the start of the test run
- Shares the same network instance across all test suites
- Maintains test isolation through separate test connectors per suite
- Properly cleans up the network after all tests complete

### New Components
- **SharedNetworkManager**: Singleton that manages the shared network lifecycle
  - Handles network launch on first request
  - Tracks references from multiple test suites
  - Ensures proper cleanup when tests complete
  
- **SharedTestSuite**: New base class for test suites
  - Replaces `BaseTestSuite` for tests that should share the network
  - Gets network instance from SharedNetworkManager
  - Only cleans up test connectors, not the network itself

- **run-shared-tests.ts**: Test runner script
  - Configures the shared network before tests run
  - Ensures proper cleanup after all tests complete
  - Uses 2-second slot time for faster test execution

### Updated Test Suites
All test suites now extend `SharedTestSuite`:
- contracts.test.ts
- datahaven-substrate.test.ts
- ethereum-basic.test.ts
- cross-chain.test.ts
- native-token-transfer.test.ts

### Package.json Updates
- `test:e2e`: Now uses shared network approach (default)
- `test:e2e:isolated`: Preserves old behavior for isolated test runs

## Benefits
- **Faster Execution**: Kurtosis and network components launch only once
- **Resource Efficiency**: Less Docker container churn and setup overhead
- **Backwards Compatible**: Original `BaseTestSuite` still available for isolated tests
- **Configurable**: Slot time set to 2 seconds for faster block production

## Testing
- [x] TypeScript compilation passes
- [ ] E2E tests pass with shared network
- [x] Tests properly share the network instance
- [x] Cleanup works correctly after all tests complete

## Breaking Changes
None - The change is backwards compatible. Tests can still use `BaseTestSuite` for isolated execution if needed.

🤖 Generated with [Claude Code](https://claude.ai/code)


test/framework/shared-network.ts
```diff
@@ -0,0 +1,165 @@
...
+  private referenceCount = 0;
+  private options: SharedNetworkOptions;
+
+  private constructor() {
+    // Default options
+    this.options = {
+      networkId: `shared-test-${Date.now()}`,
+      datahavenImageTag: "datahavenxyz/datahaven:local",
+      relayerImageTag: "datahavenxyz/snowbridge-relay:latest",
+      buildDatahaven: false,
```
gonzamontiel, line 32: we need to be careful with this default, this will prevent that the image is rebuilt when running locally. In another I added a default that depends on the environment (ci vs local), so we can leave this one. but we need to remember to change it after my PR is merged

PR #162, opened 2025-09-12
feat: added support for fisherman

undercover-cactus:
In this PR we add support for SH fisherman.


operator/node/src/service.rs
```diff
@@ -428,7 +430,7 @@ where
     // Starting StorageHub file transfer service.
     // TODO: add fisherman
```
stiiifff: TODO can be removed 😉

PR #164, opened 2025-09-12
ci: remove nextest archive

undercover-cactus:
Removing the nextest archive steps from the CI because it is taking a lot of time to download.

From latest ci runs downloading take 17min.

### DIFF
--- a/.github/workflows/task-rust-tests.yml
+++ b/.github/workflows/task-rust-tests.yml
@@ -1,9 +1,7 @@
 # Rust Tests: CI for Rust components (DataHaven runtime and node Rust tests)
 #
 # Overview:
-# 1. Prepare: This job handles the setup phase where the cargo nextest archive is created
-#    and uploaded to the workflow for use in the subsequent jobs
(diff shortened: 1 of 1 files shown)


.github/workflows/task-rust-tests.yml
```diff
@@ -13,65 +13,33 @@ on:
...
       - uses: ./.github/workflows/actions/setup-env
         with:
           cache-key: "TEST"
           install-deps: false
-      
       - name: Set build flags
         run: echo "RUSTFLAGS=${{ env.RUSTFLAGS }} -C linker=clang -C link-arg=-fuse-ld=mold" >> $GITHUB_ENV
-      
-      # Cleanup-runner skipped on self-hosted runners (bare-metal manages disk space externally)
-      - name: Install nextest
```
gonzamontiel, line 43: Wait, so nextest comes already installed, or why we skip this step if we still use `cargo-nextest` below ?

ahmadkaouk: We're still installing the nextest in the all-rust-tests job

PR #190, opened 2025-09-26
fix: fix RPC for storage hub

undercover-cactus:
In this PR we actually add the storage hub RPC config to our node in order to allow RPC call for storage hub service.

### DIFF
--- a/.github/workflows/actions/setup-env/action.yml
+++ b/.github/workflows/actions/setup-env/action.yml
@@ -98,9 +98,9 @@ runs:
           WORK_DIR="$LOCAL_DIR/tmp"
           mkdir -p "$WORK_DIR" && cd "$WORK_DIR"
           
-          # Download PostgreSQL 17 packages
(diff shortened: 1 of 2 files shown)


operator/Cargo.toml
```diff
@@ -83,7 +83,10 @@ ssz_rs_derive = { version = "0.9.0", default-features = false }
 static_assertions = { version = "1.1.0", default-features = false }
 strum = { version = "0.26.3", default-features = false, features = ["derive"] }
 strum_macros = "0.24"
-tracing = { version = "0.1.37", default-features = false }
+tracing = { version = "0.1.41", default-features = false }
```
ahmadkaouk, line 86: Changing the version in the `Cargo.toml` file is not enough, we need to also change it in the `Cargo.lock` since it's still using the old version. I think you can something like this to update the version in the Cargo.lock 
```
cargo update -p tracing-subscriber --precise 0.3.19
```

PR #191, opened 2025-09-26
feat: added the possibility to provide a config file for SH

undercover-cactus:


operator/node/src/config.rs
```diff
@@ -0,0 +1,46 @@
...
+pub fn read_config(path: &str) -> Option<Config> {
+    let path = Path::new(path);
+
+    if !path.exists() {
+        error!("Fail to find config file ({:?})", path);
+
+        return None;
+    }
+
+    let mut file = File::open(path).expect("config.toml file should exist");
```
ahmadkaouk: What is the reason for panicking here when we fail to open the file? In the same function we are treating errors as non-fatal and returning None, so why is this different?

operator/node/src/cli.rs
```diff
@@ -49,6 +49,20 @@ pub struct Cli {
...
+        "storage_layer", "storage_path", "extrinsic_retry_timeout", "sync_mode_min_blocks_behind",
+        "check_for_pending_proofs_period", "max_blocks_behind_to_catch_up_root_changes",
+        "msp_charging_period", "msp_charge_fees_task", "msp_charge_fees_min_debt",
+        "msp_move_bucket_task", "msp_move_bucket_max_try_count", "msp_move_bucket_max_tip",
+        "bsp_upload_file_task", "bsp_upload_file_max_try_count", "bsp_upload_file_max_tip",
+        "bsp_move_bucket_task", "bsp_move_bucket_grace_period",
+        "bsp_charge_fees_task", "bsp_charge_fees_min_debt",
+        "bsp_submit_proof_task", "bsp_submit_proof_max_attempts",
+    ])]
+    pub provider_config_file: Option<String>,
```
ahmadkaouk, line 65: Shouldn't passing `--provider-config-file` also conflict with the fisherman and indexer options since they will be provided by the configuration file

operator/node/src/config.rs
undercover-cactus: Yes we could remove the call to `path.exists()` and not panic when trying to `open` the file but just print the error if there one. Calling `open` will verify that the path exists.

So something like this

```rust
   let path = Path::new(path);

   let mut file = match File::open(path) {
     Ok(file) => file,
     Error(err) => {
         error!("Unable to open file : {}", err);
         return None;
     }
   };
```

operator/node/src/cli.rs
undercover-cactus: Yes it should. Fixing this.

operator/node/src/config.rs
gonzamontiel: Agree, we should have something like:
```suggestion
    let mut file = match File::open(path) {
        Ok(file) => file,
        Err(err) => {
            error!("Failed to open config file: {}", err);
            return None;
        }
    };
```

PR #216, opened 2025-10-08
misc: simplify Dockerfile to speed up build

undercover-cactus:
This PR remove the `cargo chef` step used to build the docker image used in deployment. We noticed that `cargo chef` was adding more time to the build and that removing it was saving us 40min.

Also in this PR, we removed the base image from parity which was really heavy and was filling the rest of the disk space. This broke the build. After some investigation it doesn't seem to add a lot to the build. It has been replace with the official rust image as a base to build our node.

The image used to run the image has been replaced with `debian:trixie-slim`.

In the end those changed **should not** break any of the current behavior and makes save a bit of CI time.

### DIFF
--- a/contracts/src/DataHavenServiceManager.sol
+++ b/contracts/src/DataHavenServiceManager.sol
@@ -99,9 +99,7 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
     ) external payable onlyOwner {
         // Send the new validator set message to the Snowbridge Gateway
         bytes memory message = buildNewValidatorSetMessage();
-        _snowbridgeGateway.v2_sendMessage{
-            value: msg.value
-        }(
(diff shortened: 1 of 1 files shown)


docker/datahaven-build.Dockerfile
```diff
@@ -55,30 +31,11 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     fi
 
 # --- Create final lightweight runtime image ---
-FROM docker.io/parity/base-bin:latest
```
stiiifff, line 58: Are you sure about this one ? I suppose the parity base-bin is more optimized, no ?

undercover-cactus: The base-bin is defined here : https://github.com/woss/paritytech-scripts/blob/master/dockerfiles/base-bin/Dockerfile

It doesn't do much and it using ubuntu (which is a bit heavier). Also it doesn't look like it is being updated.

PR #242, opened 2025-10-21
feat: add slashing support

undercover-cactus:
## 🔨 Add Slashing Support for Runtime

This PR introduces the slashing functionality for the DataHaven runtime, enabling punitive measures against misbehaving validators.

### Features
- Deferred slashing with configurable veto periods
- Cross-chain slashing message delivery trough Snowbridge
- Governance controls for slashing parameters and emergency cancellation

We introduced  the `external-validator-slashes` pallet, which allows to slash validators that misbehave. The slashing is triggered when an offence is reported via the offence pallet (which is already implemented). The message is sent through Snowbrige's outbound queue and the real slashing happens in the contracts side, which will come in a follow up PR.

There is a configurable window of time between the time the validator is being reported, and the time the slash is triggered. This allows that in case of an error we are still able to cancel the slashing, using a sudo account.

For convenience, we also have extrinsics for corner cases:

- **`force_inject_slash`**: Root-only function to manually inject slashes for specific validators with custom percentages. Useful for emergency situations or governance-directed slashing outside normal offence detection
- **`cancel_deferred_slash`**: Allows governance to cancel pending slashes during the defer period by specifying era and slash indices. Provides safety mechanism against false positives or malicious slash reports
- **`set_slashing_mode`**: Configurable slashing behavior with three modes - `Enabled` (normal operation), `LogOnly` (track offences without applying slashes), and `Disabled` (completely halt slashing). Critical for emergency response and testing


operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -313,6 +312,16 @@ pub mod dynamic_params {
...
+
+        #[codec(index = 35)]
+        #[allow(non_upper_case_globals)]
+        /// The Selector is the first 4 bytes of the keccak256 hash of the function signature("slashValidatorsOperator(address[])")
+        pub static SlashOperatorSelector: BoundedVec<u8, ConstU32<4>> =
+            BoundedVec::truncate_from(vec![0xca, 0x48, 0x11, 0x9f]);
+
+        #[codec(index = 36)]
+        #[allow(non_upper_case_globals)]
+        pub static DatahavenAddress: H160 = H160::repeat_byte(0x0);
```
ahmadkaouk: What is this address ? I dont see it used anywhere

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1547,6 +1547,69 @@ impl pallet_tx_pause::Config for Runtime {
...
 
+// Stub SendMessage implementation for slash pallet
+pub struct SlashesSendAdapter;
+impl pallet_external_validator_slashes::SendMessage<AccountId> for SlashesSendAdapter {
+    type Message = OutboundMessage;
+    type Ticket = OutboundMessage;
+    fn build(
+        _slashes_utils: &pallet_external_validator_slashes::SlashDataUtils<AccountId>,
+    ) -> Option<Self::Message> {
+        let calldata = Vec::new();
```
ahmadkaouk, line 1588: The calldata in the command is empty. You should extend it with the correct selector and the slash data, (you already added `SlashOperatorSelector`). As written the on-chain message will call the contract without a selector, so every slash message will fail.

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1547,6 +1547,69 @@ impl pallet_tx_pause::Config for Runtime {
...
+
+        let command = Command::CallContract {
+            target: runtime_params::dynamic_params::runtime_config::RewardsRegistryAddress::get(), // TODO: get the slash registry address
```
ahmadkaouk: Why using the `RewardsRegistryAddress`? You should add a dynamic parameter for the slash registry and use it here.

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1547,6 +1547,69 @@ impl pallet_tx_pause::Config for Runtime {
...
+            calldata,
+            gas: 1_000_000, // TODO: Determine appropriate gas value after testing
+            value: 0,
+        };
+        let message = OutboundMessage {
+            origin: runtime_params::dynamic_params::runtime_config::RewardsAgentOrigin::get(), // TODO: get the slash agent address
```
ahmadkaouk, line 1597: Same comment as above. Don't we have a Slash agent on Ethereum ?

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1547,6 +1547,69 @@ impl pallet_tx_pause::Config for Runtime {
...
+}
+
+impl pallet_external_validator_slashes::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type ValidatorId = AccountId;
+    type ValidatorIdOf = ConvertInto;
+    type SlashDeferDuration = SlashDeferDuration;
+    type BondingDuration = BondingDuration;
+    type SlashId = u32;
+    type SessionInterface = (); // FIXME
```
ahmadkaouk: Basically the implementation of `()` for the `SessionInterface` is a no-op and do nothing. We should have a proper implementation for this. We can probably try to copy what Tanssi does.

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1547,6 +1547,69 @@ impl pallet_tx_pause::Config for Runtime {
...
+    type EraIndexProvider = ExternalValidators;
+    type InvulnerablesProvider = ExternalValidators;
+    type ExternalIndexProvider = ExternalValidators;
+    type QueuedSlashesProcessedPerBlock = ConstU32<10>;
+    type WeightInfo = (); // TODO: calculate weights
+    type SendMessage = SlashesSendAdapter;
+}
+
+parameter_types! {
+    pub const SlashDeferDuration: EraIndex = polkadot_runtime_common::prod_or_fast!(0, 0);
```
ahmadkaouk, line 1639: Is it intentional to have the same value for prod and fast?  Also setting it to 0 leaves a single-era window for governance to review the slash. I wonder if it's enough time.

operator/runtime/mainnet/src/configs/runtime_params.rs
undercover-cactus: This is the address of the Datahaven address contract. It will be used to send the slashing request to it. For now we don't know the address in mainnet.

operator/runtime/mainnet/src/configs/mod.rs
undercover-cactus: We should be using the `DatahavenAddress` and not the Reward one. It is a miss. I will fix it.

operator/runtime/mainnet/src/configs/mod.rs
undercover-cactus: We can yes but this part is not working yet because it requires the AVS contract side.

operator/runtime/mainnet/src/configs/mod.rs
undercover-cactus: For now yes because we don't know which value to use. Also we don't have any governance designed to handle canceling slashing.

operator/runtime/mainnet/src/configs/mod.rs
gonzamontiel: This is a partial implementation and won't work alone. It's the runtime support for slashing.

operator/runtime/mainnet/src/configs/mod.rs
gonzamontiel: I think we should implement this @undercover-cactus

operator/runtime/mainnet/src/configs/runtime_params.rs
ahmadkaouk: Okay, so if I’m not mistaken, this is essentially the address of the Slashing Registry contract on Ethereum. If that’s the case, perhaps renaming it to `SlashingRegistryAddress` would be more intuitive.

gonzamontiel: I agree with this, and this counts for every other case, like if it's the service manager contract who handles it, then `ServiceManagerAddress` would be more accurate.

operator/runtime/mainnet/src/configs/mod.rs
undercover-cactus: I have looked at this and we don't need it because it is used when we have the staking pallets which is not our case.

operator/runtime/mainnet/src/configs/runtime_params.rs
undercover-cactus: `SlashingRegistryAddress` would not be the correct name because the idea was to directly communicate with the Datahaven AVS contract. 
`ServiceManagerAddress` is quite too abstract. Instead I propose `DatahavenAVSAddress` ?

ahmadkaouk: The AVS contract is actually called `DataHavenServiceManager`.

PR #251, opened 2025-10-25
feat: ✨ Implement dynamic fee adjustment

gonzamontiel:
# ✨ Implement Dynamic Fee Adjustment Mechanism

## Overview
Implements a dynamic fee adjustment mechanism, replacing the constant fee multiplier with an adaptive multiplier that responds to network congestion, following Moonbeam's pattern.

## Changes
- Replaces `ConstFeeMultiplier` with `TargetedFeeAdjustment` across all runtime configurations (mainnet, stagenet, testnet)
- Implements an EIP-1559-like slow-adjusting fee mechanism that prevents DoS attacks by adjusting fees based on block fullness
- **Configurable Parameters**: 
  - Target block fullness: 35%
  - Adjustment variable: 4/1000 (responds in ~1 hour at extreme congestion)
  - Two modes:
    -  `SlowAdjustingFeeUpdate` for mainnet and testnet.
    - `FastAdjustingFeeUpdate` for stagenet.
- Adds tests coverage for different fee scenarios

## Technical Details

The fee adjustment algorithm works as follows:
```
diff = (previous_block_weight - target) / maximum_block_weight
next_multiplier = prev_multiplier * (1 + (v * diff) + ((v * diff)^2 / 2))
assert(next_multiplier > min)
```
**Where:**
- `v` = AdjustmentVariable
- `target` = TargetBlockFullness  
- `min` = MinimumMultiplier

`SlowAdjustingFeeUpdate` sets a minimum multiplier of `1x` for a conservative fee adjustment, while `FastAdjustingFeeUpdate` sets it to `0.1x`, which is mainly used for dev networks / testing.


operator/runtime/testnet/tests/fee_adjustment.rs
```diff
@@ -0,0 +1,209 @@
...
+        let val50 = sim(1_000_000_000, Perbill::from_percent(50), 1);
+        let val100 = sim(1_000_000_000, Perbill::from_percent(100), 1);
+        let val600 = sim(1_000_000_000, Perbill::from_percent(50), 600);
+
+        // All values should be reasonable (within expected ranges)
+        assert!(
+            val0 > U256::from(100u128) && val0 < U256::from(100_000_000_000_000u128),
+            "Fee at 0% should be reasonable, got: {}",
+            val0
+        );
```
stiiifff, line 194: Seems like these tests are a bit "vague", no ? I mean the range of allowed values is not very strict.

gonzamontiel: Oh good catch! That was a leftover from when I was playing with different values, actually, the mainnet tests where correct, but testnet and stagenet were misaligned. The values used are similar from [moonbeam's tests](https://github.com/moonbeam-foundation/moonbeam/blob/8254fc33aa6744872bd5fb191f2118b935b75ff5/runtime/moonbase/tests/integration_test.rs#L3078-L3079)). Should be fixed now.

PR #253, opened 2025-10-27
fix: add explicit sovereign account balance check in unlock_tokens

ahmadkaouk:
Add defensive validation to ensure the Ethereum sovereign account has sufficient balance before unlocking tokens. This addresses an audit finding where the lack of explicit balance checking created an unreliable security control that depended on implicit runtime behavior.

Changes:
- Add InsufficientSovereignBalance error variant for clear error messaging
- Add explicit balance check in unlock_tokens before transfer
- Update tests across all runtimes (testnet, stagenet, mainnet) to validate the specific error is returned when sovereign account has insufficient funds

The explicit check provides better error messages that can propagate through the Ethereum bridge and makes debugging sovereign account balance issues easier.


operator/pallets/datahaven-native-transfer/src/lib.rs
```diff
@@ -275,13 +277,14 @@ pub mod pallet {
         ///
         /// Transfers tokens from the Ethereum sovereign account back to user
         pub fn unlock_tokens(who: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+            let sovereign = T::EthereumSovereignAccount::get();
+            let balance = T::Currency::balance(&sovereign);
+
+            // Explicit defensive check for sufficient sovereign balance
+            ensure!(balance >= amount, Error::<T>::InsufficientSovereignBalance);
```
gonzamontiel: I think this would conflict if we allow an ExistentialDeposit different from 0 in a future. We should contemplate that corner case.

ahmadkaouk: Currency::transfer already safeguards the sovereign account—it refuses to execute if the send would take the balance below the existential deposit—so we wouldn’t lose funds even if the runtime raises ExistentialDeposit later. I added a check that subtracts ED (minimum_balance()) before unlocking as an extra measure to surface the`InsufficientSovereignBalance` error instead of the generic TokenError.

PR #264, opened 2025-10-29
feat: set slashing mode in genesis config

undercover-cactus:
In this PR we set the slashing mode value in the genesis config. For the 3 different runtime we specify the slashing mode : `mainnet/testnet` is set to `Disabled` and for `stagenet`  to `LogOnly`.


operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -235,6 +237,22 @@ pub mod pallet {
     #[pallet::storage]
     pub type SlashingMode<T: Config> = StorageValue<_, SlashingModeOption, ValueQuery>;
 
+    #[pallet::genesis_config]
+    #[derive(frame_support::DefaultNoBound)]
+    pub struct GenesisConfig<T: Config> {
+        // Slashing mode
+        pub slashing_mode: SlashingModeOption,
+        #[serde(skip)]
```
stiiifff, line 245: Not sure this is needed .. other pallets don't need to depend on serde.

undercover-cactus: Yes it is needed. The other pallets use primitives struct that have the Serialize/Deserialize trait already.

PR #267, opened 2025-10-30
fix: add timestamp safe mode

gonzamontiel:
Following up https://github.com/datahaven-xyz/datahaven/pull/265, we also need to add Timestamp to the whitelisted Runtime calls.

- [x] Add `RuntimeCall::Timestamp` to `SafeModeWhitelistedCalls`
- [x] Add safe mode test to check it produces blocks

### DIFF
--- a/test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
+++ b/test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
@@ -31,6 +31,7 @@ describeSuite({
 
       expect(enteredUntil.isSome, "Safe mode should be active").to.be.true;
     });
+
     it({
       id: "T01",
       title: "should produce blocks while in safe mode",
@@ -50,10 +51,17 @@ describeSuite({
           "Blocks should continue to be produced in safe mode"
         );
 
+        const exitBlockBefore = await getSubstrateBlockNumber();
         const exitSafeModeCall = api.tx.safeMode.forceExit();
         const exitSudoTx = api.tx.sudo.sudo(exitSafeModeCall);
 
         await context.createBlock(exitSudoTx);
+
+        // Verify the exit block was created (ensures state is updated)
+        const exitBlockAfter = await getSubstrateBlockNumber();
+        expect(exitBlockAfter, "Exit block should have been created").to.be.greaterThan(
+          exitBlockBefore
+        );
 
         const enteredUntilAfterExit = (await api.query.safeMode.enteredUntil()) as any;
         expect(!enteredUntilAfterExit.isSome, "Safe mode should be deactivated").to.be.true;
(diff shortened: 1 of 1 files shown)


test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        expect(enteredUntil.isSome, "Safe mode should be active").to.be.true;
+
+        const blockBefore = await getSubstrateBlockNumber();
+
+        // System calls are whitelisted, so this should succeed
+        const remarkTx = api.tx.system.remark("0x74657374"); // "test" in hex
+
+        // Create block with the remark transaction
+        await context.createBlock(remarkTx);
```
ahmadkaouk: test can pass even if system.remark is blocked, because it only asserts that a block was produced, not that the remark extrinsic executed successfully

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
```diff
@@ -0,0 +1,151 @@
...
+        // Verify safe mode is active
+        expect(enteredUntil.isSome, "Safe mode should be active").to.be.true;
+
+        // Create a block - this implicitly includes timestamp extrinsic
+        // The fact that this succeeds proves Timestamp is whitelisted
+        await context.createBlock();
+
+        // Verify the block was created (contains valid timestamp)
+        const block = await context.viem().getBlock({ blockTag: "latest" });
+        expect(Number(block.timestamp)).to.be.greaterThan(0);
```
ahmadkaouk, line 87: A better approach would be to remember the pre-test timestamp and assert the new block’s timestamp increased or that the timestamp extrinsic appears in the events.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        if (!enteredUntil.isSome) {
+          const enterSafeModeCall = api.tx.safeMode.forceEnter();
+          const sudoTx = api.tx.sudo.sudo(enterSafeModeCall);
+          await context.createBlock(sudoTx);
+
+          enteredUntil = (await api.query.safeMode.enteredUntil()) as any;
+        }
+
+        expect(enteredUntil.isSome, "Safe mode should be active").to.be.true;
```
ahmadkaouk: Since this is done for each test you can move the “ensure safe mode is active” block into a `beforeEach` hook.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
gonzamontiel: You are right, I'll work in a better testing scenario in a different PR, we are mainly interested in that blocks were produced, but agree this test doesn't prove whitelisted calls.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
gonzamontiel: Same as above.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
gonzamontiel: Done

PR #272, opened 2025-10-31
feat: add Slashing mode has a runtime configurable parameter

undercover-cactus:


operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -362,6 +366,11 @@ pub mod pallet {
             let processed = Self::process_slashes_queue(T::QueuedSlashesProcessedPerBlock::get());
             T::WeightInfo::process_slashes_queue(processed)
         }
+
+        fn on_runtime_upgrade() -> Weight {
+            SlashingMode::<T>::put(T::SlashingMode::get());
+            T::DbWeight::get().reads(1)
```
undercover-cactus: Really not sure about the weight that should be returned here. What we do in tis function is pretty simple and costless.

ahmadkaouk: You are writing to storage so you should do
```suggestion
            T::DbWeight::get().writes(1)
```

PR #275, opened 2025-11-03
fix: safe mode test

gonzamontiel:
## Fix safe mode test
- Fix test isolation so safe mode is exited gracefully and doesn't affect other tests execution
- Refactored test for readability adding a helper to check for events in block
- Added some test scenarios entering and exiting safe mode


test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
```diff
@@ -78,34 +109,99 @@ describeSuite({
...
+      test: async () => {
+        const currentBlock = await getSubstrateBlockNumber();
+        const scheduleAtBlock = currentBlock + 10;
+
+        const call = api.tx.system.remarkWithEvent("0xabcd");
+        const scheduleTx = api.tx.scheduler.schedule(scheduleAtBlock, null, 0, call);
+
+        const blockHash = await context.createBlock(scheduleTx);
+
+        const extrinsicSuccess = await checkEvent("system.ExtrinsicSuccess", blockHash);
```
ahmadkaouk: `scheduler.schedule` requires a `Root` origin `type ScheduleOrigin = EnsureRoot<AccountId>` but it's submitted as a plain extrinsic, so It will fail.  Because the test only checks for a generic `system.ExtrinsicSuccess` event—which is already emitted by the timestamp inherent each block—it will report a pass even though the scheduler call never succeeded. Please wrap the schedule call in sudo and assert on a scheduler-specific success signal (e.g. `scheduler.Scheduled`) or the returned execution result.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
```diff
@@ -78,34 +109,99 @@ describeSuite({
...
 
-        // Exit safe mode
-        const exitBlockBefore = await getSubstrateBlockNumber();
-        const exitSafeModeCall = api.tx.safeMode.forceExit();
-        const exitSudoTx = api.tx.sudo.sudo(exitSafeModeCall);
+        const blockHash = await context.createBlock(remarkTx);
 
-        await context.createBlock(exitSudoTx);
+        const remarkExecuted = await checkEvent("system.Remarked", blockHash);
+        const extrinsicSuccess = await checkEvent("system.ExtrinsicSuccess", blockHash);
```
ahmadkaouk, line 132: `checkEvent("system.ExtrinsicSuccess", blockHash)` will always pass even if the remark extrinsic fail since `checkEvent` checks for a generic system.ExtrinsicSuccess event—which is already emitted by the timestamp inherent each block

gonzamontiel: Yeah, true, but we are checking both events, also `system.Remarked`. I can remove the ExtrinsicSuccess if you think is redundant, I think it's okay.

ahmadkaouk: Simply highlighting that `checkEvent(“system.ExtrinsicSuccess”)` doesn’t necessarily confirm that the remark tx has succeeded. Checking `system.Remarked` is more than sufficient in this case.

test/datahaven/suites/dev/common/test-block/test-block-safe-mode.ts
gonzamontiel: Done.

PR #287, opened 2025-11-07
feat: 🚀 add storage-hub nodes to CLI

gonzamontiel:
##  Add StorageHub nodes to CLI

This PR adds StorageHub node infrastructure to the CLI and fixes CLI flag handling and improves the CLI logic a bit.

### Fix: CLI safeguards
- Prevents contract deployment and validator operations when `--ndc` flag is used
- Skips Kurtosis service display in summary when `--nlk` flag is used

### Feat: Dockerized Storage Hub Nodes
- Adds 5 Docker containers: PostgreSQL, MSP, BSP, Indexer, and Fisherman nodes
- New CLI flags: `--storagehub` `--no-storagehub` to control StorageHub nodes launch
- Automatic provider funding and registration (Charleth for MSP and Dorothy for BSP)
- Exposes nodes on ports 9945-9948 for local development

**TODO**
- [x] MSP & BSP associated pre-funded account.
- [x] Call `forceMspSignUp` and `forceBspSignUp` extrinsics


test/cli/handlers/launch/index.ts
```diff
@@ -78,15 +78,17 @@ const launchFunction = async (options: LaunchOptions, launchedNetwork: LaunchedN
...
-    parameterCollection
-  });
-
-  await performValidatorOperations(options, launchedNetwork.elRpcUrl, contractsDeployed);
+  if (options.deployContracts) {
+    const contractsDeployed = await deployContracts({
+      rpcUrl: launchedNetwork.elRpcUrl,
+      verified: options.verified,
+      blockscoutBackendUrl,
+      deployContracts: options.deployContracts,
```
undercover-cactus, line 86: If we are checking before calling the function so we don't need to pass it as an argument to the `deployContracts` after. I am ok with this change but we need to make sure that there is no hidden behavior inside the `deployContracts` function

test/cli/handlers/launch/kurtosis.ts
```diff
@@ -34,7 +34,6 @@ export const launchKurtosis = async (
   if (!shouldLaunchKurtosis) {
     logger.info("👍 Skipping Kurtosis Ethereum network launch. Done!");
 
-    await registerServices(launchedNetwork, options.kurtosisEnclaveName);
```
undercover-cactus, line 37: Why are we removing this ?

test/cli/handlers/launch/storagehub.ts
```diff
@@ -1,36 +1,103 @@
...
-  await deployStorageHubComponents(deployOptions as any, launchedNetwork);
+  // Check if we're in local Docker mode or K8s deploy mode
+  if (launchedNetwork.networkId === NETWORK_ID) {
+    // LOCAL DOCKER MODE (CLI launch)
+    await launchStorageHubDocker(options, launchedNetwork);
+  } else {
+    // KUBERNETES MODE (deploy command)
+    const deployOptions = {
+      environment: "local" as const,
+      skipStorageHub: !options.storagehub,
```
undercover-cactus, line 46: I know it is not your change but I am noticing that we have this weird behavior where we change the option to `skipStorageHub` we should stay consistent and maybe refactor this in another PR.

test/cli/handlers/launch/kurtosis.ts
gonzamontiel: Because if we are skipping kurtosis, then we don't need to register kurtosis services.

test/cli/handlers/launch/index.ts
gonzamontiel: Well, I didn't want to do a full refactor _yet_. The reason behind the change is that `launchedNetwork.elRpcUrl` is an invariant and will fail when calling `deployContracts()`. So I did an extra check. Inside the function there's another check that also has an interactive check, I don't like that. Personally I would remove all that and have the CLI work by `profiles` where you define the profiles based on the different use cases and let the command do everything without much prompting.

undercover-cactus: I agree with not going crazy on the refactoring for now. If it works then I approve it.

test/scripts/register-providers.ts
```diff
@@ -0,0 +1,168 @@
...
+    logger.debug(`MSP ID: ${mspId}`);
+
+    const mspCall = typedApi.tx.Providers.force_msp_sign_up({
+      who: PROVIDERS.msp.accountId,
+      msp_id: mspId,
+      capacity: PROVIDERS.msp.capacity,
+      multiaddresses: [],
+      value_prop_price_per_giga_unit_of_data_per_block: BigInt(1000),
+      commitment: Binary.fromText(`msp-${PROVIDERS.msp.name.toLowerCase()}`),
+      value_prop_max_data_limit: BigInt(1_000_000),
```
stiiifff: You can specify 1Gb here (with 10Gb capacity).

test/scripts/register-providers.ts
```diff
@@ -0,0 +1,168 @@
...
+    logger.info(`Registering MSP (${PROVIDERS.msp.name})...`);
+    const mspId = generateProviderId(PROVIDERS.msp.accountId);
+    logger.debug(`MSP ID: ${mspId}`);
+
+    const mspCall = typedApi.tx.Providers.force_msp_sign_up({
+      who: PROVIDERS.msp.accountId,
+      msp_id: mspId,
+      capacity: PROVIDERS.msp.capacity,
+      multiaddresses: [],
+      value_prop_price_per_giga_unit_of_data_per_block: BigInt(1000),
```
stiiifff: Let's use `18520000000` which is the value we used for Stagenet & Testnet. This value corresponds to the amount of HAVE (18 decimals) at a price of 0.5$, for storing 1Gb of data per block if we target a cost of 0.2$ / 50GB / month.

test/scripts/register-providers.ts
```diff
@@ -0,0 +1,168 @@
...
+    // Register MSP
+    logger.info(`Registering MSP (${PROVIDERS.msp.name})...`);
+    const mspId = generateProviderId(PROVIDERS.msp.accountId);
+    logger.debug(`MSP ID: ${mspId}`);
+
+    const mspCall = typedApi.tx.Providers.force_msp_sign_up({
+      who: PROVIDERS.msp.accountId,
+      msp_id: mspId,
+      capacity: PROVIDERS.msp.capacity,
+      multiaddresses: [],
```
stiiifff: The multi-address of the provider is needed.

test/scripts/register-providers.ts
gonzamontiel: Done.

test/scripts/register-providers.ts
gonzamontiel: Done.

test/scripts/register-providers.ts
gonzamontiel: Done.

PR #292, opened 2025-11-10
feat: statically build binary

undercover-cactus:
## Summary

This pr build add a new statically build the pibpq dependecnies into the binary to the release assets. It also add a job that will verify that we can still build statically the libpq into the binary when `Cargo.toml` or `Cargo.lock` is being modified. It is useful to catch failing build early and just not when we want to do a new release.

The new binary can be used without having to install postgres libraries on the system and should work as is.

## What changed

- Added a new feature `static` that allow to build the libpq dependencies
- Added a new task in the CI that build the static binary every time we changed some dependencies to make sure the build is still successful
-  Add the static binary as an assets for release


operator/.cargo/config.toml
```diff
@@ -3,4 +3,10 @@ lto = "thin"
...
 [registries.crates-io]
-protocol = "sparse" 
\ No newline at end of file
+protocol = "sparse" 
+
+[target.x86_64-unknown-linux-gnu]
+rustflags = [
+    "-C", "link-arg=-s",
+    "-C", "target-feature=-crt-static",
+]
```
stiiifff: Can we only enable this for a separate `production-static` profile ?

undercover-cactus: Will we still be using this profile in the CI ? When would we use one profile or the other ?

stiiifff: I think the main use case for a statically compiled binary is for production. That being said, I'm not comfortable to enable this for `production` profile yet .. let's have a separate profile to test this.

undercover-cactus: Couldn't do it in a profile but it is behind a feature now ( e.g `cargo build --release --locked --features fast-runtime,static`).
I have added a task to build it only when adding new dependencies to make sure it doesn't break it.

.github/workflows/task-build-static-operator.yml
```diff
@@ -0,0 +1,46 @@
+name: DataHaven Operator Static Build
+
+on:
+  pull_request:
+    paths:
+      # Only check the build if we are adding a dependency to save runner time
+      - 'Cargo.toml'
+      - 'Cargo.lock'
```
ahmadkaouk: Why this is overly restrictive? it won't run on Rust code changes, for example.  Also, I think you should update the path to `operator/Cargo.toml` and `operator/Cargo.lock`

operator/Cargo.toml
```diff
@@ -304,6 +304,10 @@ shp-types = { git = "https://github.com/Moonsong-Labs/storage-hub.git", tag = "v
 ## Precompiles
 pallet-evm-precompile-file-system = { git = "https://github.com/Moonsong-Labs/storage-hub.git", tag = "v0.1.4", default-features = false }
 
+# Static linking
+#### Needed to build static binaries ####
+pq-sys = { version = "0.7.2" }
```
ahmadkaouk: There is a version mismatch between what is specified here `0.7.2` and what is used in Cargo.lock `0.7.4`.

.github/workflows/actions/setup-env/action.yml
```diff
@@ -68,9 +73,16 @@ runs:
...
+    - name: Install Protoc
+      if: inputs.install-deps == 'true'
+      uses: arduino/setup-protoc@v3
+      with:
+        repo-token: ${{ github.token }}
+
+    # Auto-install missing dependencies when install-deps is false (for self-hosted runners) can be skipped when statically build the node
     - name: Setup system dependencies (self-hosted)
-      if: inputs.install-deps == 'false'
+      if: inputs.install-deps == 'false' && inputs.skip-libpq == 'false'
```
ahmadkaouk, line 85: Setting `skip-libpq` to true will skip install LLVM/Clang block part and a lot more which I guess is still needed.  I think we should divide the step in two one to setup up the system dependencies for the self hosted runners and one for libpq.

Also, I know it's not related to this PR , but `install-deps` naming is misleading.

.github/workflows/task-build-static-operator.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set build flags
+        run: echo "RUSTFLAGS=${{ env.RUSTFLAGS }} -C linker=clang -C link-arg=-fuse-ld=mold" >> $GITHUB_ENV
+
+      - name: Build node binary
+        run: |
+          cargo build --release --locked --features fast-runtime,static
+
+      - name: Test binary
+        run: |
+          ldd ./target/release/datahaven-node
```
ahmadkaouk, line 44: This is not really testing if the static link succeeded but only prints ldd output

.github/workflows/task-build-static-operator.yml
```diff
@@ -0,0 +1,46 @@
...
+          cache-key: BUILD-RELEASE
+          install-deps: false
+          skip-libpq: true
+
+      - name: Set build flags
+        run: echo "RUSTFLAGS=${{ env.RUSTFLAGS }} -C linker=clang -C link-arg=-fuse-ld=mold" >> $GITHUB_ENV
+
+      - name: Build node binary
+        run: |
+          cargo build --release --locked --features fast-runtime,static
```
ahmadkaouk, line 40: is this binary only for testing purposes ? If not, we should not enable the `fast-runtime` feature.

.github/workflows/task-build-static-operator.yml
```diff
@@ -0,0 +1,46 @@
+name: DataHaven Operator Static Build
```
ahmadkaouk, line 1: What is this job used for? Is it only to verify that static build compiles successfully ? I don't see the binary produced, uploaded, or used anywhere yet.

.github/workflows/task-build-static-operator.yml
undercover-cactus: Yes good catch.

.github/workflows/task-build-static-operator.yml
undercover-cactus: The `ldd` call only show the dynamic linking. It is more informative. We are testing that the binary is correctly build and can be executes the line after `./target/release/datahaven-node --version`.

.github/workflows/task-build-static-operator.yml
undercover-cactus: This job is only testing that the build is working and that there is not a lib issue. The build of the actual binary is happening when creating a release. This job is only here to catch early compilation issue and not when we are doing the release.

This is why we only build it when there is a change in Cargo.toml or Cargo.lock

.github/workflows/task-build-static-operator.yml
undercover-cactus: Yes, this job only make sure that we can build the binary statically successfully. It is only being ran when we update dependencies to catch issue early.

The asset is being build and uploaded with the other assets on release.

.github/workflows/actions/setup-env/action.yml
undercover-cactus: I can confirm that we don't need it. It also seems to me that LLVM is already available on the server. I am not sure for clang.
See  : https://github.com/datahaven-xyz/datahaven/actions/runs/19572049272/job/56047676981?pr=292

I agree the `install-deps` could be better. We could have an array of the dependencies we want passed to this action.

gonzamontiel: Agree with having an array of deps, we can leave that for a follow up PR.

PR #295, opened 2025-11-11
feat: injecting contracts feature for e2e testing

undercover-cactus:
In this PR, we introduce a way to save Ethereum state into a file. This saved state can then be injected into Ethereum to speed up e2e initial test setup.

This is a rewrite of the now closed PR https://github.com/datahaven-xyz/datahaven/pull/90 .

It uses a an external tool written in rust to save state from the Ethereum running container : https://github.com/undercover-cactus/Chaos


test/cli/handlers/launch/index.ts
```diff
@@ -96,7 +96,7 @@ const launchFunction = async (options: LaunchOptions, launchedNetwork: LaunchedN
 
   await launchRelayers(options, launchedNetwork);
 
-  await launchStorageHubComponents(options, launchedNetwork);
+  // await launchStorageHubComponents(options, launchedNetwork);
```
stiiifff: Why comment this, there is a CLI flag that allows to skip it I think.

test/Makefile
```diff
@@ -0,0 +1,8 @@
+version := v0.1.2
+
+generate:
```
stiiifff: This should be named with a more descriptive name.

test/launcher/network/index.ts
```diff
@@ -203,24 +204,79 @@ export const launchNetwork = async (
...
-      verified: options.verified ?? false,
-      blockscoutBackendUrl,
-      parameterCollection
-    });
+    // await deployContracts({
+    //   rpcUrl: launchedNetwork.elRpcUrl,
+    //   verified: options.verified ?? false,
+    //   blockscoutBackendUrl,
+    //   parameterCollection
+    // });
```
stiiifff: Cleanup unused code.

test/launcher/network/index.ts
```diff
@@ -203,24 +204,79 @@ export const launchNetwork = async (
...
-    // 5. Setup validators
-    logger.info("🔐 Setting up validators...");
-    await setupValidators({
-      rpcUrl: launchedNetwork.elRpcUrl
-    });
+    // // 5. Setup validators
+    // logger.info("🔐 Setting up validators...");
+    // await setupValidators({
+    //   rpcUrl: launchedNetwork.elRpcUrl
+    // });
```
stiiifff: Cleanup unused code.

test/cli/handlers/launch/index.ts
undercover-cactus: You can skip it using with a flag but it is not compatible when using `--all`. The function fail when calling it from the `launch` command.
I will re-add it.

test/launcher/network/index.ts
```diff
@@ -186,41 +190,79 @@ export const launchNetwork = async (
...
     await fundValidators({
       rpcUrl: launchedNetwork.elRpcUrl
     });
 
-    // 5. Setup validators
-    logger.info("🔐 Setting up validators...");
-    await setupValidators({
-      rpcUrl: launchedNetwork.elRpcUrl
-    });
+    // We are injecting contracts but we still need the addresses
```
gonzamontiel, line 214: can you improve this comment ? this addreses are needed for parameter collection, and since deployContracts() is not being called anymore, we need to call this chunk of code.

test/launcher/network/index.ts
```diff
@@ -186,41 +189,27 @@ export const launchNetwork = async (
...
+    await updateParameters(parameterCollection);
```
ahmadkaouk, line 215: `launchNetwork` now always skips `deployContracts` and still calls `updateParameters`, but the deployment artifacts it reads (contracts/deployments/anvil.json) are git‑ignored and are no longer created anywhere in the flow. On a fresh clone those files don’t exist, so `parseDeploymentsFile` throws and the collection falls back to the template with all null values.

test/scripts/deploy-contracts.ts
```diff
@@ -40,6 +40,8 @@ export const buildContracts = async () => {
     stdout: buildStdout
   } = await $`forge build`.cwd("../contracts").nothrow().quiet();
 
+  await $`bun run biome format --files-max-size=2000000 --write ../contracts/deployments/state-diff.json`;
+
```
ahmadkaouk: Why are we actually running this here? Also we don’t guard the call, so if `state-diff.json` is missing, the Biome CLI exits with a non-zero status, and `deployContracts` will fail.

test/launcher/network/index.ts
```diff
@@ -186,41 +189,27 @@ export const launchNetwork = async (
         kurtosisEnclaveName: kurtosisEnclaveName,
         blockscout: options.blockscout ?? false,
         slotTime: options.slotTime || 2,
-        kurtosisNetworkArgs: options.kurtosisNetworkArgs
+        kurtosisNetworkArgs: options.kurtosisNetworkArgs,
+        injectContracts: true // Forcing it to be true to run e2e tests
```
ahmadkaouk, line 190: The `injectContracts` flag shouldn’t be hard-coded to `true` in `launchNetwork` or at least we should be able to configure for each test suite. Different test suites may require opting in or out of contract injection. Additionally, since contract injection is currently only supported on Linux, we won't be able to run tests on Mac.

What you can do  is:
1. Add `injectContracts?: boolean` to `NetworkLaunchOptions` in `test/launcher/types/index.ts`
2. Update `launchNetwork` to use `options.injectContracts ?? true` instead of hard-coded `true`
3. Add `injectContracts?: boolean` to `TestSuiteOptions["networkOptions"]` in `test/framework/suite.ts`
```typescript
export class NativeTokenTransferTestSuite extends BaseTestSuite {
  constructor() {
    super({
      suiteName: "native-token-transfer",
      networkOptions: {
        injectContracts: false,  // Disable snapshot injection
        slotTime: 2
      }
    });
  }
```
This should work I guess.

undercover-cactus: Injecting contracts works on both Linux and MacOS so it won't be any issue.

The snapshot is just the deployment of the smart contract so I don't really see any cases where you wouldn't use it. Not injecting the contracts in the test suite would defy the purpose of this feature.

We can still add an option in an other PR as this has been already delayed for months.

test/scripts/deploy-contracts.ts
undercover-cactus: It was added here in the original PR.

test/launcher/network/index.ts
ahmadkaouk: If `injectContracts` stays hard-coded there’s no practical way to iterate on contract changes: every tweak to the Solidity code would require me to regenerate a fresh snapshot (bun cli launch --all, make generate-ethereum-state, stop everything, rerun the suite) before I can even run tests, which completely defeats the “faster test loop” goal. We need to be able to disable injection so we can deploy re-run e2e tests immediately—otherwise  contracts edits turn into multiple CLI launches + snapshot rebuilds just to validate the change.

I’m perfectly fine with leaving this to another PR.  it’s not an urgent matter to implement.

undercover-cactus: Fair. I think what we need is to be able to pass argument to test using the CLI to be able to override. For example : 
```
$ bun test suites/rewards-message.test.ts --no-inject-contracts
```
So it makes it possible to develop and run the suite test. And to be sure that a new state has been generated to set a git hook to check if files has been modified in the contracts folder to be sure that tests run with the latest contracts injected.

ahmadkaouk: Yes, this would be even better if we could do it.

gonzamontiel: This follow up PR we are talking about I started to work on, but should have some priority since we didn't test the contract injection enough to see that every case is working, we only rely on our e2e tests.

gonzamontiel (approved):
I approve with the same concerns that @ahmadkaouk remarked, I still think we need to test this further. Having said that, I already started working on a follow up PR to make the contract injection optional, and some other checks that are missing in this PR.

PR #304, opened 2025-11-18
feat: ✨ Implement inflation mechanism for validator rewards

stiiifff:
## Summary

This PR introduces a configurable inflation system for validator rewards with an annual target rate and optional treasury allocation.

### Inflation Mechanism
- **Annual inflation rate runtime parameter**: Set to 5% default
- **EraInflationProvider**: Calculates per-era inflation based on total issuance and annual rate
  - Formula: `per_era_inflation = (total_issuance × annual_rate) / eras_per_year`

### Treasury Allocation
- **InflationTreasuryProportion parameter**: Set to 20% default
- **ExternalRewardsInflationHandler**: Mints inflation and distributes between:
  - 80% to rewards account (for validator rewards)
  - 20% to treasury account
- Treasury receives allocation via `mul_floor()`, with remainder going to rewards to ensure no tokens lost to rounding

### Runtime Integration
- Configured across all three runtimes: mainnet, testnet, and stagenet
- Consistent parameters across all environments

### Testing
- Updated all tests to account for 80/20 split between rewards and treasury
- Added precision tolerance (±1 unit) for Perbill rounding edge cases


operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -339,6 +339,24 @@ pub mod dynamic_params {
...
+        /// Default: 10% per annum
+        /// This rate is divided across all eras in a year to calculate per-era inflation.
+        pub static InflationTargetedAnnualRate: Perbill = Perbill::from_percent(10);
+
+        #[codec(index = 38)]
+        #[allow(non_upper_case_globals)]
+        /// Proportion of inflation rewards allocated to the treasury.
+        /// Default: 20% of minted rewards go to treasury, 80% to validator rewards
+        /// The treasury portion is minted separately and sent to the treasury account.
+        pub static InflationTreasuryProportion: Perbill = Perbill::from_percent(20);
```
ahmadkaouk, line 357: I wonder if we should set the treasury proportion to 100% at the beginning. Initially, we want to run whitelisted validators, and I believe they won’t be rewarded. We can change it once we allow for external validators.

operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -339,6 +339,24 @@ pub mod dynamic_params {
...
         pub static DatahavenAVSAddress: H160 = H160::repeat_byte(0x0);
+
+        // ╔══════════════════════ Validator Rewards Inflation ═══════════════════════╗
+
+        #[codec(index = 37)]
+        #[allow(non_upper_case_globals)]
+        /// Targeted annual inflation rate.
+        /// Default: 10% per annum
+        /// This rate is divided across all eras in a year to calculate per-era inflation.
+        pub static InflationTargetedAnnualRate: Perbill = Perbill::from_percent(10);
```
ahmadkaouk: The inflation rate is set to 10%, but the PR description mentions 5%.

stiiifff: Forgot to push the commit 😅

operator/runtime/mainnet/src/configs/runtime_params.rs
stiiifff: Still to be defined. But the idea could be to accumulate some funds, ahead of time, to be later transferred to the AVS for the operator rewards.

ahmadkaouk: Okay, that makes sense.

PR #306, opened 2025-11-18
feat: ✨ Performance-Based Validator Rewards and Inflation Scaling

stiiifff:
## Summary

Building on #304, this PR implements two complementary mechanisms to improve validator incentives and network performance:

1. **Performance-Based Validator Rewards** (session-level)
2. **Inflation Scaling** (era-level)

### Old Model (main branch) vs New Model

| Metric | Old Model (20 pts/block) | New Model (320 pts/block pool) |
|--------|--------------------------|--------------------------------|
| **Per Block** | Author: 20 pts, Others: 0 | Author: ~196 pts, Others: ~4 pts each |
| **Formula** | Direct author reward | 60% authoring + 30% liveness + 10% base |
| **Per Session** (600 blocks, 32 validators) | 12,000 total pts | 192,000 total pts |
| **Per Validator/Session** (uniform) | ~375 pts | ~6,000 pts |
| **Per Validator/Era** (6 sessions) | ~2,250 pts | ~36,000 pts |
| **Offline Validator** | 0 pts | ~600 pts/session (base only) |
| **Over-performer (150% blocks)** | 150% of fair share | Up to 130% reward (soft cap) |

### Key Differences
- **Pool-based**: New model adds 320 points to a shared pool per block, distributed via formula
- **Liveness rewarded**: 30% of rewards go to validators who are online (heartbeat OR block authorship)
- **Base guarantee**: 10% ensures all active validators receive minimum rewards
- **Soft cap**: Prevents extreme over-performance rewards (max 150% of fair share credited)

## Performance-Based Validator Rewards

Introduces a **60/30/10 reward formula** that rewards validators based on their contribution during each session:

- **60%** based on block production (with soft cap allowing up to 150% of fair share)
- **30%** based on liveness (ImOnline heartbeat OR block authorship)
- **10%** guaranteed base reward for all active validators

### Key Features
- Tracks individual validator block authorship per session
- Calculates fair share dynamically: `fair_share = total_blocks / total_validator_count`
- Fair share uses **total** validator count (including whitelisted) since all validators occupy block slots
- **Soft cap**: Over-performers can earn credit up to 150% of their fair share (configurable via `OperatorRewardsFairShareCap` at 50%)
- With 60% BlockAuthoringWeight, this gives over-performers up to **30% bonus reward**
- **BasePointsPerBlock**: Defines points added to pool per block produced (default: 320)
- Integrates with SessionManager for automatic point awards at session end
- Excludes whitelisted validators from rewards (but includes them in fair share calculation)
- Slashing check disabled but hook retained for future use
- Points accumulate across sessions within an era

### Dynamic Parameters (Governance-Adjustable)
- `OperatorRewardsBlockAuthoringWeight`: Weight for block authoring (default: 60%)
- `OperatorRewardsLivenessWeight`: Weight for liveness (default: 30%)
- `OperatorRewardsFairShareCap`: Soft cap percentage above fair share (default: 50%)

## Inflation Scaling

Implements **dynamic inflation scaling** that adjusts total inflation based on network block production:

- **Minimum**: 20% of base inflation (network halt protection)
- **Maximum**: 100% of base inflation (caps at expected blocks)
- **Linear scaling** between minimum and maximum based on performance

### Scaling Examples
- 0% blocks produced → 20% inflation (safety floor)
- 50% blocks produced → 60% inflation  
- 100% blocks produced → 100% inflation
- >100% blocks produced → capped at 100%

### Configuration
- **ExpectedBlocksPerEra**: Computed as `SessionsPerEra × EpochDurationInBlocks`
- **MinInflationPercent**: 20%
- **MaxInflationPercent**: 100%

## Combined Effect

These mechanisms work together to create a comprehensive incentive structure:

1. **Session rewards** encourage individual validator performance and uptime
2. **Era inflation scaling** incentivizes collective network health
3. **Minimum inflation floor** protects against network halt
4. **Soft cap** allows over-performers to earn up to 30% bonus while preventing extreme centralization

### Pallet Changes
- Add `BlocksAuthoredInSession` storage for per-validator tracking
- Add `BlocksProducedInEra` storage for total network tracking (cleaned up with HistoryDepth)
- Add `note_block_author()` function called on block production
- Add `award_session_performance_points()` function with configurable 60/30/10 formula
- Add `calculate_scaled_inflation()` function for era-level scaling
- Update `on_era_end()` to use scaled inflation
- Integrate with SessionManager via wrapper types
- Defensive weight validation: proportionally scales if sum > 100%

### Configuration Parameters
- `ValidatorSet`: Provides active validator list
- `LivenessCheck`: Uses `ImOnline::is_online()` (heartbeat OR block authorship)
- `SlashingCheck`: Integration with slashing pallet (currently disabled)
- `BasePointsPerBlock`: Points added to pool per block (default: 320)
- `BlockAuthoringWeight`: Dynamic parameter (60%)
- `LivenessWeight`: Dynamic parameter (30%)
- `FairShareCap`: Dynamic parameter (50%)
- `ExpectedBlocksPerEra`: Computed from session/epoch config
- `MinInflationPercent`: 20%
- `MaxInflationPercent`: 100%

### Runtime Updates
- Full configuration added to mainnet, testnet, and stagenet runtimes
- Dynamic parameters added to `runtime_params.rs` for governance control
- Uses `prod_or_fast!()` macro for environment-specific parameters
- `ValidatorIsOnline` uses `ImOnline::is_online()` for accurate liveness detection

## Testing

- **76 tests passing** ✅
- Comprehensive coverage of both mechanisms

### Test Coverage
- Inflation scaling at 0%, 25%, 50%, 75%, 100%, >100% blocks
- Session performance with 60/30/10 formula
- Fair share calculations with soft cap (150%)
- Whitelisted validator exclusion from rewards (with correct fair share using total count)
- Total points verification (sum of individual = total)
- Whitelisted over-producer scenarios
- Overflow protection (large block counts, near-u32::MAX)
- End-to-end session to era flow
- MockLivenessCheck mirrors ImOnline behavior (block authorship = online)
- Multiple eras with different performance levels
- Edge cases (zero participation, single validator, large numbers)
- BlocksProducedInEra cleanup on era start

### Reward Distribution

Previously, rewards were distributed equally among all validators regardless of their contribution. Now:

- **Performance-based**: Validators earn rewards proportional to their block production (60%), liveness (30%), and a guaranteed base (10%)
- **Pool-based**: `BasePointsPerBlock` defines points added to pool per block (320), distributed via formula
- **Fair share uses total validators**: Ensures non-whitelisted aren't penalized for whitelisted validators' block slots
- **Soft cap**: Block production rewards allow up to 150% of fair share (50% cap = 30% bonus with 60% weight)
- **Slashing check disabled**: Hook retained for future use, but currently not applied

### Inflation Mechanism

Previously, the full calculated inflation was minted each era. Now:

- **Scaled by performance**: Total inflation scales between 20%-100% based on actual blocks produced vs expected
- **Safety floor**: Even with zero blocks, 20% of inflation is still minted to prevent complete halt
- **Network incentive**: Collective block production directly impacts total rewards available

### Pallet Configuration

The `pallet-external-validators-rewards` Config now requires additional types:
- `BlockAuthoringWeight`, `LivenessWeight`, `FairShareCap` for reward formula
- `ValidatorSet`, `LivenessCheck`, `SlashingCheck` for validator tracking
- `ExpectedBlocksPerEra`, `MinInflationPercent`, `MaxInflationPercent` for inflation scaling


operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -206,6 +249,19 @@ pub mod pallet {
...
+    #[pallet::storage]
+    #[pallet::unbounded]
+    pub type BlocksAuthoredInSession<T: Config> =
+        StorageMap<_, Twox64Concat, T::AccountId, u32, ValueQuery>;
+
+    /// Track the total number of blocks produced in each era.
+    /// Used to scale inflation based on network performance.
+    #[pallet::storage]
+    pub type BlocksProducedInEra<T: Config> =
+        StorageMap<_, Twox64Concat, EraIndex, u32, ValueQuery>;
```
ahmadkaouk, line 287: We should probably add `#[pallet::storage]` for this storage item to. Also, it would be good to add a periodic cleanup that remove old entries in `on_era_start`, matching what we do currently for `RewardPointsForEra`. This prevents unbounded growth by keeping only HistoryDepth eras.

operator/runtime/testnet/src/configs/mod.rs
```diff
@@ -1514,6 +1504,46 @@ impl pallet_external_validators_rewards::types::SendMessage for RewardsSendAdapt
...
+        let validators = Session::validators();
+        if let Some(index) = validators.iter().position(|v| v == account) {
+            // Check if this validator sent a heartbeat (not just produced blocks)
+            ImOnline::received_heartbeat_in_current_session(index as u32)
+        } else {
+            // Not a validator in current session, consider offline
+            false
+        }
+    }
+}
```
ahmadkaouk, line 1532: If I understand correctly with the current implementation if a validator authored blocks but their heartbeat was never received/included, they’ll be treated as offline for this check. Is it intended ? like if a validator produced blocks, but somehow missed the heartbeat shouldn't they still be considered online ?

operator/runtime/stagenet/src/configs/mod.rs
```diff
@@ -1513,6 +1503,46 @@ impl pallet_external_validators_rewards::types::SendMessage for RewardsSendAdapt
...
+        )
+    }
+}
+
+parameter_types! {
+    /// Expected number of blocks per era for inflation scaling
+    /// Calculation: SessionsPerEra × EpochDurationInBlocks
+    /// - Production: 6 sessions × 600 blocks = 3600 blocks (6 hours at 6s/block)
+    /// - Fast runtime: 1 session × 10 blocks = 10 blocks (for testing)
+    pub const ExpectedBlocksPerEra: u32 = polkadot_runtime_common::prod_or_fast!(3600, 10);
```
ahmadkaouk: We should use the constants instead of the hardcoded values
```suggestion
    /// Expected number of blocks per era for inflation scaling
    /// Calculation: SessionsPerEra × EpochDurationInBlocks
    pub const ExpectedBlocksPerEra: u32 =
        (SessionsPerEra::get() as u32).saturating_mul(EpochDurationInBlocks::get() as u32);
```

operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -272,6 +328,212 @@ pub mod pallet {
...
+            }
+
+            // Fair share: expected blocks per validator using checked math
+            let expected_blocks_per_validator = if total_blocks > 0 {
+                total_blocks
+                    .checked_div(active_validator_count)
+                    .unwrap_or(1)
+            } else {
+                1 // Default to 1 when no blocks were produced
+            };
```
ahmadkaouk: `expected_blocks_per_validator` can be zero when `total_blocks` > 0 but `total_blocks` < `active_validator_count`. We should ensure a minimum of 1.
```suggestion
            // Fair share: expected blocks per validator using checked math
            let expected_blocks_per_validator = total_blocks
                .checked_div(active_validator_count)
                .unwrap_or(1)
                .max(1);
```

operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -272,6 +328,212 @@ pub mod pallet {
...
+        /// other reward mechanisms.
+        pub fn award_session_performance_points(
+            session_index: SessionIndex,
+            validators: Vec<T::AccountId>,
+            whitelisted_validators: Vec<T::AccountId>,
+        ) {
+            // Calculate total blocks for the session
+            let total_blocks: u32 = BlocksAuthoredInSession::<T>::iter()
+                .map(|(_, count)| count)
+                .sum();
```
ahmadkaouk, line 470: I wonder if we should exclude blocks that are authored by whitelisted validators. Whitelisted validators can inflate the baseline while not competing for rewards, making it harder for non-whitelisted validators to achieve good scores. An example of that:
```
10 validators (2 whitelisted, 8 non-whitelisted)
100 total blocks (whitelisted produced 50, non-whitelisted produced 50)
Current: expected = 100/10 = 10 blocks
Non-whitelisted validator with 6 blocks → block_score = 60%
```
If we exclude whitelisted:
```
total_blocks = 50 (only non-whitelisted)
validator_count = 8 (only non-whitelisted)
expected = 50/8 ≈ 6 blocks
Same validator with 6 blocks → block_score = 100%
```

operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -272,6 +328,212 @@ pub mod pallet {
...
+                    Perbill::one()
+                } else {
+                    Perbill::zero()
+                };
+
+                // Apply 50/30/20 weighting formula
+                let weighted_score = Perbill::from_percent(50)
+                    .saturating_mul(block_score)
+                    .saturating_add(Perbill::from_percent(30).saturating_mul(liveness_score))
+                    .saturating_add(Perbill::from_percent(20)); // 20% base guarantee
```
ahmadkaouk: I would suggest making the weights (50, 30, 20) configurable instead of using hard-coded constants since:
1. We can benefit from using different weights in different networks
2. It would be useful to adjust the weights without a runtime upgrade if you use dynamic parameters

operator/runtime/stagenet/src/configs/mod.rs
```diff
@@ -1513,6 +1503,46 @@ impl pallet_external_validators_rewards::types::SendMessage for RewardsSendAdapt
...
+
+/// Wrapper to check if a validator has been slashed in a given era
+pub struct ValidatorSlashChecker;
+impl pallet_external_validators_rewards::SlashingCheck<AccountId> for ValidatorSlashChecker {
+    fn is_slashed(era_index: u32, validator: &AccountId) -> bool {
+        pallet_external_validator_slashes::ValidatorSlashInEra::<Runtime>::contains_key(
+            era_index, validator,
+        )
+    }
+}
```
ahmadkaouk, line 1538: This check does not take into consideration the deferrall period when the slash can be cancelled via `cancel_deferred_slash()`. So Rewards are nullified immediately even during the deferral period.

operator/runtime/testnet/src/configs/mod.rs
stiiifff: I specifically wanted to dissociate block production from ImOnline heartbeats. The only reason I see (correct me if you see other cases) for missing a heartbeat over the duration of a 1-hour session is a misconfiguration of the ImOnline session key, or not passing the `--offchain-worker` flag when running the validator (but the default value is `when-authority` so it's unlikely).

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Yeah I thought about it but then it will significantly scale down the inflation in the early phase of the network when the validator set will likely be composed mainly of whitelisted validators.

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Agreed

operator/runtime/testnet/src/configs/mod.rs
ahmadkaouk: Yes correct, I know that it’s highly unlikely to happen, especially if the validator is producing blocks. But, considering the worst-case scenario, if the heartbeat is somehow missed due to the off-chain worker crashing or if the transaction isn’t included in a block in time, the validator would lose 30% of the reward which is a lot.

operator/runtime/stagenet/src/configs/mod.rs
stiiifff: It's been disabled for now, but the hook is retained for future use.

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Implemented as suggested 👍

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: My bad, my previous comment was for inflation. You're right, I updated the logic to only take the regular validators into account when calculating the fair share.

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Fixed

operator/runtime/stagenet/src/configs/mod.rs
stiiifff: Fixed

operator/runtime/testnet/src/configs/mod.rs
stiiifff: Agreed. Reverted back to using the ImOnline pallet's `is_online` function which takes both heartbeat and block authoring (at least 1 block durning current session) into account.

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Fixed

stiiifff:
## Latest Changes (16a18964)

**Increased FairShareCap from 20% to 50%**

This change allows over-performing validators to earn credit for up to 150% of their fair share (previously 120%).

### Rationale
With the 60% BlockAuthoringWeight, the previous 20% cap only gave over-performers a 12% bonus reward (60% × 20%). The new 50% cap provides a more meaningful **30% bonus** (60% × 50%) for validators who produce more blocks than their fair share.

### Changes
- `OperatorRewardsFairShareCap`: 20% → 50% (all runtimes)
- Updated all related comments and documentation
- Fixed remaining "50/30/20" formula references to "60/30/10"
- Renamed `test_weighting_formula_50_30_20` → `test_weighting_formula_60_30_10`

All 56 tests passing ✅

stiiifff:
## Latest Changes

### Commit `9f93500f` - Parameter Rename and Fair Share Fix

**Parameter Rename:**
- Renamed `AuthorBaseRewardPoints` → `BasePointsPerBlock` for clarity
- Updated default value from 20 to 320 points per block for easier mental model:
  - Per block: author gets ~196 pts (60%), each non-author gets ~4 pts (40%/31)
  - Per session (600 blocks, 32 validators): ~6,000 pts/validator
  - Per era (6 sessions): ~36,000 pts/validator

**Fair Share Calculation Fix:**
- Fixed `fair_share` to use **total validator count** (not just non-whitelisted)
- Rationale: Whitelisted validators still occupy block production slots, so their slots shouldn't inflate fair share expectations for non-whitelisted validators
- Example: 12 validators (10 whitelisted, 2 non-whitelisted), 600 blocks → fair_share = 50 (not 300)

**Liveness/Base Pool Fix:**
- Fixed pool division to use total validator count for consistent reward distribution
- Ensures non-whitelisted validators get their proportional share regardless of whitelist composition

### Commit `d0ca6fc9` - Test Improvements

**MockLivenessCheck Fix:**
- Now mirrors real ImOnline behavior: validators who author blocks are considered online even without heartbeat
- This matches production behavior where block authorship proves liveness

**New Test Coverage (76 tests total):**
- Total points verification (sum check)
- Whitelisted over-producer scenarios (critical edge case)
- Overflow protection tests
- End-to-end session → era flow tests

All tests passing ✅

operator/runtime/testnet/src/configs/mod.rs
gonzamontiel: shouldn't we remove this now ? in all three runtimes

operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -272,6 +352,305 @@ pub mod pallet {
...
+        /// Each validator receives points based on configurable weights (default 60/30/10):
+        /// - **BlockAuthoringWeight**: Block production score with soft cap allowing over-performance
+        /// - **LivenessWeight**: Liveness score (1.0 if online, 0.0 otherwise)
+        /// - **Base guarantee**: Remainder (100% - block - liveness) always awarded
+        ///
+        /// Final points = BASE_POINTS × (block_weight × block_score + liveness_weight × liveness_score + base_weight)
+        ///
+        /// # Block Production Scoring
+        ///
+        /// - Fair share = total_blocks / non_whitelisted_validators
```
gonzamontiel: This has changed to use total validator count, so the description is outdated.

```suggestion
        /// - Fair share = total_blocks / total_validator_count
```

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1524,6 +1515,47 @@ impl pallet_external_validators_rewards::types::SendMessage for RewardsSendAdapt
...
+        pallet_external_validator_slashes::ValidatorSlashInEra::<Runtime>::contains_key(
+            era_index, validator,
+        )
+    }
+}
+
+parameter_types! {
+    /// Expected number of blocks per era for inflation scaling.
+    /// Computed as SessionsPerEra × EpochDurationInBlocks to ensure consistency.
+    pub ExpectedBlocksPerEra: u32 = (SessionsPerEra::get() as u32)
```
gonzamontiel, line 1548: Are we expecting that at some point this value is different from: sessions per era * blocks per session ? I say this because to ensure consistency we can directly compute this value inside the pallet. We expect to tweak this value at some point to be _independent of time_ ?

PS: is a bit misleading (or making an implicit assumption) the fact that here we treat `EpochDurationInBlocks` the same as `SessionDurationInBlocks`.

stiiifff:
@gonzamontiel You're right re. the incorrect `~128 pts/session (base only)` value, I've updated the PR description.

operator/runtime/mainnet/src/configs/mod.rs
stiiifff: It can vary per runtime, and whether it's a fast runtime or not.

operator/pallets/external-validators-rewards/src/lib.rs
stiiifff: Good catch ! Fixed in https://github.com/datahaven-xyz/datahaven/pull/306/commits/8798a9b745fd33091127350afc6419a7a96cf9b7

operator/runtime/testnet/src/configs/mod.rs
stiiifff: You're right, fixed in https://github.com/datahaven-xyz/datahaven/pull/306/commits/8798a9b745fd33091127350afc6419a7a96cf9b7

PR #308, opened 2025-11-19
fix: 🔨 Replace FreezeChainOnFailedMigration with SafeMode handler

stiiifff:
## Summary

Replaced the `DefaultFailedMigrationHandler` (which completely froze the chain on migration failures) with `EnterSafeModeOnFailedMigration` across all three runtimes (mainnet, stagenet, testnet). When a migration fails, the chain now automatically enters SafeMode instead of freezing, allowing governance to intervene and fix issues while preventing regular user transactions.

## Problem

Previously, when a runtime migration failed, the chain would use `FreezeChainOnFailedMigration`, which completely halted all operations including governance functions. This made it impossible to recover from migration failures without manual intervention at the node level.

## Solution

Implemented `EnterSafeModeOnFailedMigration` which:
- **Enters SafeMode** when a migration fails: the chain remains _indefinitely_ under safe mode until it is disabled, either with Sudo or Governance.
- **Allows governance operations** to continue (Sudo, SafeMode, TxPause, Preimage, Scheduler, etc.)
- **Blocks regular user transactions** to prevent interaction with potentially inconsistent storage
- **Falls back to freezing** if SafeMode cannot be entered

### Core Implementation
- **`runtime/common/src/migrations.rs`**: Added `FailedMigrationHandler<SafeMode>` type alias that wraps `EnterSafeModeOnFailedMigration` with comprehensive documentation
- **All three runtimes** (`mainnet`, `stagenet`, `testnet`):
  - Updated `pallet_migrations::Config::FailedMigrationHandler` to use `FailedMigrationHandler<SafeMode>`
  - Removed obsolete TODO comments

### Tests
Added comprehensive migration failure tests to all three runtimes:
- **`failed_migration_enters_safe_mode`**: Verifies SafeMode is activated, expiry is set, and event is emitted
- **`safe_mode_allows_governance_during_migration_failure`**: Confirms governance can exit SafeMode after migration failure
- **`migrations_force_calls_are_root_only`**: Existing test for migration management permissions

### DIFF
--- a/operator/runtime/mainnet/src/configs/mod.rs
+++ b/operator/runtime/mainnet/src/configs/mod.rs
@@ -511,7 +511,8 @@ impl pallet_beefy::Config for Runtime {
     type AncestryHelper = BeefyMmrLeaf;
     type WeightInfo = ();
     type KeyOwnerProof = <Historical as KeyOwnerProofSystem<(KeyTypeId, BeefyId)>>::Proof;
-    type EquivocationReportSystem = ();
+    type EquivocationReportSystem =
+        pallet_beefy::EquivocationReportSystem<Self, Offences, Historical, ReportLongevity>;
 }
 
 parameter_types! {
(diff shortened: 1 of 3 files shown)


ahmadkaouk (approved):
LGTM!  Just one thing `frame_support::migrations::EnterSafeModeOnFailedMigration` set duration to Bounded::max_value(). That means `SafeMode` entered after a migration failure will persist indefinitely and not just for one day as the PR description state.

PR #309, opened 2025-11-19
feat: ✨ Add datahaven native transfer precompile

gonzamontiel:
## DataHaven Native Transfer Precompile

Implements EVM precompile at address `0x00000000000000000000000000000007F5` (2073) to expose `pallet-datahaven-native-transfer` functionality to the EVM layer.

### Features
- **Transfer to Ethereum**: Locks native tokens and sends them via Snowbridge to Ethereum addresses
- **View Functions**: Query paused state, total locked balance, and sovereign account address

### Implementation
- Precompile using `#[precompile_utils::precompile]` macro with proper gas accounting
- 15+ test cases covering success/failure scenarios
- Solidity interface with NatSpec documentation for contract integration

Enables seamless cross-chain transfers of DataHaven native tokens to Ethereum L1.

### DIFF
--- a/operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
+++ b/operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0-only
+    // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity >=0.8.3;
 
 /// @dev The DataHavenNativeTransfer precompile address.
@@ -18,22 +18,11 @@ interface DataHavenNativeTransfer {
     /// @param amount The amount of tokens locked
     event TokensLocked(address indexed account, uint256 amount);
 
-    /// @notice Emitted when tokens are unlocked from Ethereum
-    /// @param account The account that received unlocked tokens
-    /// @param amount The amount of tokens unlocked
-    event TokensUnlocked(address indexed account, uint256 amount);
-
     /// @notice Emitted when tokens are transferred to Ethereum
     /// @param from The account initiating the transfer
     /// @param to The Ethereum address receiving the tokens
     /// @param amount The amount of tokens transferred
     event TokensTransferredToEthereum(address indexed from, address indexed to, uint256 amount);
-
-    /// @notice Emitted when the pallet is paused
-    event Paused();
-
-    /// @notice Emitted when the pallet is unpaused
-    event Unpaused();
 
     /// @notice Transfer DataHaven native tokens to Ethereum
     /// @dev Locks tokens in the sovereign account and sends message through Snowbridge
--- a/operator/precompiles/datahaven-native-transfer/src/lib.rs
+++ b/operator/precompiles/datahaven-native-transfer/src/lib.rs
@@ -33,6 +33,15 @@ use sp_core::{H160, U256};
 use sp_runtime::traits::Dispatchable;
 use sp_std::marker::PhantomData;
 
+/// Solidity selector for the TokensLocked event:
+/// keccak256("TokensLocked(address,uint256)")
+pub const SELECTOR_LOG_TOKENS_LOCKED: [u8; 32] = keccak256!("TokensLocked(address,uint256)");
+
+/// Solidity selector for the TokensTransferredToEthereum event:
+/// keccak256("TokensTransferredToEthereum(address,address,uint256)")
+pub const SELECTOR_LOG_TOKENS_TRANSFERRED_TO_ETHEREUM: [u8; 32] =
+    keccak256!("TokensTransferredToEthereum(address,address,uint256)");
+
 #[cfg(test)]
 mod mock;
 #[cfg(test)]
@@ -102,6 +111,12 @@ where
             return Err(revert("Fee must be greater than zero"));
         }
 
+        // Reserve gas for emitting the two EVM logs we produce on success:
+        // - TokensLocked(address,uint256)  -> 2 topics
+        // - TokensTransferredToEthereum(address,address,uint256) -> 3 topics
+        handle.record_log_costs_manual(2, 32)?;
+        handle.record_log_costs_manual(3, 32)?;
+
         // Build the call
         let call = NativeTransferCall::<Runtime>::transfer_to_ethereum {
             recipient: recipient_h160,
@@ -112,6 +127,25 @@ where
 
         // Dispatch the call - this will handle gas costs and error reporting
         RuntimeHelper::<Runtime>::try_dispatch(handle, Some(caller).into(), call, 0)?;
+
+        // Emit EVM log mirroring the TokensLocked pallet event
+        log2(
+            handle.context().address,
+            SELECTOR_LOG_TOKENS_LOCKED,
+            handle.context().caller,
+            solidity::encode_event_data(amount),
+        )
+        .record(handle)?;
+
+        // Emit EVM log for the high-level transfer intent to Ethereum
+        log3(
+            handle.context().address,
+            SELECTOR_LOG_TOKENS_TRANSFERRED_TO_ETHEREUM,
+            handle.context().caller,
+            recipient_h160,
+            solidity::encode_event_data(amount),
+        )
+        .record(handle)?;
 
         Ok(())
     }


stiiifff:
As discussed, not sure the Pause/Unpause operations should be exposed in the precompile. Otherwise, looks good.

operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity >=0.8.3;
+
+/// @dev The DataHavenNativeTransfer precompile address.
+address constant DATAHAVEN_NATIVE_TRANSFER_ADDRESS = 0x00000000000000000000000000000000000007f5;
```
ahmadkaouk: `0x00000000000000000000000000000000000007F5` corresponds to 2037 which is not the address of the precompile used in the runtime.
https://github.com/datahaven-xyz/datahaven/blob/b76e13d911e346d0279af1d9a3af99e9169edf00/operator/runtime/mainnet/src/precompiles.rs#L145-L149

operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
```diff
@@ -0,0 +1,70 @@
...
+
+/// @dev The DataHavenNativeTransfer precompile instance.
+DataHavenNativeTransfer constant DATAHAVEN_NATIVE_TRANSFER_CONTRACT =
+    DataHavenNativeTransfer(DATAHAVEN_NATIVE_TRANSFER_ADDRESS);
+
+/// @author The DataHaven Team
+/// @title DataHaven Native Transfer Interface
+/// @notice Interface for transferring DataHaven native tokens to/from Ethereum via Snowbridge
+/// @custom:address 0x00000000000000000000000000000000000007F5
```
ahmadkaouk: Should fix the address

operator/precompiles/datahaven-native-transfer/README.md
```diff
@@ -0,0 +1,286 @@
...
+This precompile exposes the `pallet-datahaven-native-transfer` functionality to the EVM layer, allowing smart contracts to transfer DataHaven native tokens to Ethereum via Snowbridge.
+
+## Overview
+
+The DataHaven Native Transfer precompile provides an EVM-compatible interface for:
+- Transferring native tokens from DataHaven to Ethereum
+- Managing the pallet's operational state (pause/unpause)
+- Querying transfer statistics and system state
+
+**Precompile Address:** `0x00000000000000000000000000000000000007F5` (2073 decimal)
```
ahmadkaouk: Fix the address

operator/precompiles/datahaven-native-transfer/src/tests.rs
```diff
@@ -0,0 +1,759 @@
...
+#[test]
+fn test_view_functions_gas_costs() {
+    ExtBuilder::default()
+        .with_balances(vec![(EthereumSovereign.into(), 1000)])
+        .build()
+        .execute_with(|| {
+            // isPaused should have minimal gas cost
+            precompiles()
+                .prepare_test(Alice, precompile_address(), PCall::is_paused {})
+                .expect_cost(0) // TODO: Calculate actual expected cost
```
ahmadkaouk: What’s the deal with this TODO?

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        handle.record_cost(RuntimeHelper::<Runtime>::db_write_gas_cost())?;
+
+        // Convert caller address to substrate account
+        let caller = Runtime::AddressMapping::into_account_id(handle.context().caller);
+
+        // Build the call
+        let call = NativeTransferCall::<Runtime>::pause {}.into();
+
+        // Dispatch the call - this will fail if caller doesn't have permission
+        RuntimeHelper::<Runtime>::try_dispatch(handle, Some(caller).into(), call, 0)?;
```
ahmadkaouk: Thinking again about this, this call will always revert. The precompile dispatches with a signed origin while the runtime requires `EnsureRoot` for `PauseOrigin`. No EVM caller can produce Root as far as I know, so the always revert with `BadOrigin`.

It's probably better to remove the pause and unpause calls.

operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
```diff
@@ -0,0 +1,70 @@
...
+    /// @param from The account initiating the transfer
+    /// @param to The Ethereum address receiving the tokens
+    /// @param amount The amount of tokens transferred
+    event TokensTransferredToEthereum(address indexed from, address indexed to, uint256 amount);
+
+    /// @notice Emitted when the pallet is paused
+    event Paused();
+
+    /// @notice Emitted when the pallet is unpaused
+    event Unpaused();
```
ahmadkaouk: We are declaring events here but the precompile Rust implementation never constructs a log or calls record_log. So no EVM receipt logs are produced.

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// - `amount`: Amount of tokens to transfer (in smallest unit)
+    /// - `fee`: Fee to incentivize relayers (in smallest unit)
+    #[precompile::public("transferToEthereum(address,uint256,uint256)")]
+    fn transfer_to_ethereum(
+        handle: &mut impl PrecompileHandle,
+        recipient: Address,
+        amount: U256,
+        fee: U256,
+    ) -> EvmResult {
+        // Ensure we're not in a static context
```
ahmadkaouk: What does this comment mean?

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        handle.record_cost(RuntimeHelper::<Runtime>::db_read_gas_cost())?;
```
ahmadkaouk: Why are we recording a db_read_gas_cost here?

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /// Unpause the pallet
+    ///
+    /// Allows token transfers again after being paused.
+    /// Only callable by accounts with the PauseOrigin permission.
+    #[precompile::public("unpause()")]
+    fn unpause(handle: &mut impl PrecompileHandle) -> EvmResult {
+        // Record db write cost
+        handle.record_cost(RuntimeHelper::<Runtime>::db_write_gas_cost())?;
```
ahmadkaouk: why we are recording a db_write_gas_cost here ?

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+#[cfg(test)]
+mod tests;
+
+type BalanceOf<Runtime> =
+    <<Runtime as pallet_datahaven_native_transfer::Config>::Currency as Inspect<
+        <Runtime as frame_system::Config>::AccountId,
+    >>::Balance;
+
+/// Storage read size for Paused storage value (bool = 1 byte + overhead)
+pub const PAUSED_STORAGE_SIZE: usize = 32;
```
ahmadkaouk: It's actually 33.

```
  - Key = pallet prefix (Twox128, 16 B) + storage prefix (Twox128, 16 B) → 32 bytes.
  - Value = SCALE‐encoded bool → 1 byte.
  - Raw key+value payload = 32 + 1 = 33 bytes.
```

operator/precompiles/datahaven-native-transfer/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+            .map_err(|_| RevertReason::custom("Fee overflow").in_field("fee"))?;
+
+        // Validate amounts are non-zero
+        if amount_balance.into() == U256::zero() {
+            return Err(revert("Amount must be greater than zero"));
+        }
+
+        if fee_balance.into() == U256::zero() {
+            return Err(revert("Fee must be greater than zero"));
+        }
```
ahmadkaouk, line 112: These checks are redundant. The pallet already enforces these (and more). It’s not harmful and can save a bit of gas on early failure, but it does create two sources of truth: if pallet validation changes, the precompile would still reject.

gonzamontiel: Yeah I thought of that, but shouldn't add tests be enough for maintaning consistency ?

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: I think this was misunderstood. The value is used for computing the db read cost, which for a boolean should be 1 byte. This value was being over conservative by trying to read a whole db word of 32 bytes. 
Looking at other precompile examples in Moonbeam (like https://github.com/moonbeam-foundation/moonbeam/blob/ff24ffea8649ed91fff837f3f674b28cc6aefb5c/precompiles/balances-erc20/src/lib.rs#L196) they seem to compute only for the SCALE-encoded value, in this case 1 byte. So I'll change it in the code to be 1 byte.

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: Removed.

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: my bad, left over from testing

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: also a leftover, is not necessary, since try_dispatch will account already for storage reads and we don't perform additional storage reads here

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: removed.

operator/precompiles/datahaven-native-transfer/src/lib.rs
ahmadkaouk: We actually account for the storage prefix and pallet prefix when computing proof sizes for storage reads. To verify a Merkle proof for a storage read, we need:
- The full storage key (to locate the node in the state trie): `twox128(pallet_name) ++ twox128(storage_name) ++ hasher(key)`
- The value (the actual data)

Without the pallet and storage prefixes, we couldn't reconstruct the full key and verify the proof is valid. Here is a valid example:
https://github.com/datahaven-xyz/datahaven/blob/0618e84268c82e4a794ccfbc2308d3af786aa9a1/operator/precompiles/collective/src/lib.rs#L42-L44

Speaking of this, since we’ve now set the `GasLimitPovSizeRatio` to 0, this value doesn’t hold much importance anymore.

operator/precompiles/datahaven-native-transfer/src/lib.rs
gonzamontiel: So, we already have tests that revert in this corner cases (zero address and amounts), the overflow check is necessary for the try_into(), so I think we should keep them as they don't validate some specific logic, but corner cases (transfer a zero amount is never a use case)

operator/precompiles/datahaven-native-transfer/DataHavenNativeTransfer.sol
stiiifff: @gonzamontiel has this comment been addressed ?

gonzamontiel: Adressed now. I deleted the events we cannot record the log for, and added the record log for TockensLocked and TokensTransferredToEthereum

PR #315, opened 2025-11-21
test: only inject contracts in e2e tests if INJECT_CONTRACTS env is 'true'

undercover-cactus:
In this PR we add an environment variable `INJECT_CONTRACTS`. This environment variable specify if the contracts should be injected in the e2e tests. By default it is false. The environment variable is set to `true` in the CI job that run the e2e tests.

We are using a environment variable because `bun test` doesn't allow for passing extra arguments.

A note about the new variable has been added in the documentation to inform about the new behavior.

### DIFF
--- a/test/launcher/network/index.ts
+++ b/test/launcher/network/index.ts
@@ -1,8 +1,8 @@
 import { $ } from "bun";
 import { getContainersMatchingImage, getPortFromKurtosis, logger } from "utils";
 import { ParameterCollection } from "utils/parameters";
-import { deployContracts } from "../contracts";
 import { updateParameters } from "../../scripts/deploy-contracts";
+import { deployContracts } from "../contracts";
 import { launchLocalDataHavenSolochain } from "../datahaven";
 import { getRunningKurtosisEnclaves, launchKurtosisNetwork } from "../kurtosis";
 import { setDataHavenParameters } from "../parameters";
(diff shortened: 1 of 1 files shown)


contracts/deployments/anvil-rewards-info.json
```diff
@@ -1,5 +1 @@
-{
-    "RewardsAgent": "0xac06641381166cf085281c45292147f833C622d7",
-    "RewardsAgentOrigin": "0x0000000000000000000000000000000000000000000000000000000000000000",
-    "updateRewardsMerkleRootSelector": "0xdc3d04ec"
-}
\ No newline at end of file
+{"RewardsAgent": "0xac06641381166cf085281c45292147f833C622d7","RewardsAgentOrigin": "0x0000000000000000000000000000000000000000000000000000000000000000","updateRewardsMerkleRootSelector": "0xdc3d04ec"}
```
gonzamontiel: Was this intentional? I would avoid committing unrelated style changes. Also this could generate unnecessary conflicts.

ahmadkaouk: I believe this is auto-formatting for JSON files, the old style is much better and more readable.

gonzamontiel: Yup, agree, that's why I asked

undercover-cactus: Yep my bad I am removing it.

gonzamontiel (approved):

stiiifff (approved):
LGTM 🚀

PR #326, opened 2025-12-01
test : improve contract injection

undercover-cactus:
## Summary

This PR improve the generating state workflow. It will also check for outdated state-diff.json and add a practical script to easily generate a new one.

The way we generate state has also been changed to make it work with macOS M1 system. We don't run the tool in the container anymore but instead directly on the machine.

## What changes

* A check-generated-state.js script was added to quickly look for outdated test
* The check was added in the CI
* A generate-contracts.ts script was added to easily generate the new state with the new instructions to run on MacOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,44 +1,73 @@
 # DataHaven 🫎
 
-An EVM-compatible Substrate blockchain secured by EigenLayer, bridging Ethereum and Substrate ecosystems through trustless cross-chain communication.
+AI-First Decentralized Storage secured by EigenLayer — a verifiable storage network for AI training data, machine learning models, and Web3 applications.
 
 ## Overview
 
-DataHaven is an EigenLayer Actively Validated Service (AVS) that combines:
+DataHaven is a decentralized storage and retrieval network designed for applications that need verifiable, production-scale data storage. Built on [StorageHub](https://github.com/Moonsong-Labs/storage-hub) and secured by EigenLayer's restaking protocol, DataHaven separates storage from verification: providers store data off-chain while cryptographic commitments are anchored on-chain for tamper-evident verification.
 
-- **EVM Compatibility**: Full Ethereum support via Frontier pallets for smart contracts and dApps
-- **EigenLayer Security**: Validator set secured by Ethereum's economic security through restaking
-- **Cross-chain Bridge**: Seamless asset and message transfers with Ethereum via Snowbridge
-- **Dynamic Validators**: Operator registry managed on-chain through EigenLayer contracts
-- **Performance Rewards**: Validator incentives distributed cross-chain from Ethereum
+**Core Capabilities:**
+
+- **Verifiable Storage**: Files are chunked, hashed into Merkle trees, and committed on-chain — enabling cryptographic proof that data hasn't been tampered with
+- **Provider Network**: Main Storage Providers (MSPs) serve data with competitive offerings, while Backup Storage Providers (BSPs) ensure redundancy through decentralized replication with on-chain slashing for failed proof challenges
+- **EigenLayer Security**: Validator set secured by Ethereum restaking — DataHaven validators register as EigenLayer operators with slashing for misbehavior
+- **EVM Compatibility**: Full Ethereum support via Frontier pallets for smart contracts and familiar Web3 tooling
+- **Cross-chain Bridge**: Native, trustless bridging with Ethereum via Snowbridge for tokens and messages
 
 ## Architecture
 
-DataHaven bridges two major blockchain ecosystems:
+DataHaven combines EigenLayer's shared security with StorageHub's decentralized storage infrastructure:
 
 ```
-┌───────────────────────────────────────────────────────────────┐
-│                       Ethereum (L1)                           │
-│  ┌────────────────────────────────────────────────────────┐   │
-│  │  EigenLayer AVS Contracts                              │   │
-│  │  • DataHavenServiceManager (operator lifecycle)        │   │
-│  │  • RewardsRegistry (performance tracking)              │   │
-│  └────────────────────────────────────────────────────────┘   │
-│                            ↕                                  │
-│                  Snowbridge Protocol                          │
-└───────────────────────────────────────────────────────────────┘
-                             ↕
-┌───────────────────────────────────────────────────────────────┐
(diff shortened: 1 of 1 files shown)


test/docs/state-diff-validation.md
ahmadkaouk, line 1: This doc seems to describe a proposed plan, not the implemented solution. I guess we can remove it.

test/scripts/check-generated-state.js
```diff
@@ -0,0 +1,20 @@
...
+
+for (const content of contents) {
+  const stats = fs.lstatSync(path.join(contractsPath, content));
+  if (stats.isFile()) {
+    // if this a .sol file check the date
+    if (stateGeneratedCTime < stats.mtime) {
+      throw "State generated file is outdated. Please regenerate it with `bun ./scripts/generate-contracts.ts`";
+    }
+  }
+}
```
ahmadkaouk: THe script doesn't provide any feedback when everything is up-to-date. We can add a success message at the end of the script.

test/scripts/check-generated-state.js
```diff
@@ -0,0 +1,20 @@
+const fs = require("node:fs");
+const path = require("node:path");
+
+const contractsPath = "../contracts/src";
+
+// Get our generated state creation time
+const stateGeneratedCTime = fs.lstatSync("../contracts/deployments/state-diff.json").ctime;
```
ahmadkaouk: In CI environments, git checkout creates all files freshly. The timestamps (mtime/ctime) will reflect the time of the checkout, not the author date or commit date. So comparing timestamps is not reliable

test/scripts/test-generated-contracts.ts
```diff
@@ -0,0 +1,53 @@
...
+      throw new Error(
+        "state-diff.json checksum mismatch!\n\n" +
+          `Expected: ${expectedChecksum}\n` +
+          `Actual:   ${actualChecksum}\n\n` +
+          "This means the contracts have changed. Please run 'bun generate:contracts' to regenerate state-diff.json"
+      );
+    }
+
+    expect(actualChecksum).toBe(expectedChecksum);
+  });
```
ahmadkaouk: This only proves that `state-diff.json` hasn't been modified independently of its checksum file. It does not prove that `state-diff.json` is up-to-date with the contract sources. If a user modifies `Contract.sol` but doesn't run `generate:contracts`, this test will still pass (because the old json and old checksum still match), but the state is stale.

test/Makefile
```diff
@@ -1,8 +1,16 @@
 version := v0.1.2
 
+# generate-ethereum-state:
+# #	docker exec $(shell docker ps -aq --filter name='el-1-reth') bash -c "apt update && apt install -y wget"
+# #	docker exec $(shell docker ps -aq --filter name='el-1-reth') bash -c "wget https://github.com/undercover-cactus/Chaos/releases/download/${version}/chaos-linux-amd64-${version}.tar.gz"
+# 	docker cp chaos-linux-amd64-${version}.tar.gz $(shell docker ps -aq --filter name='el-1-reth'):chaos-linux-amd64-${version}.tar.gz
+# 	docker exec $(shell docker ps -aq --filter name='el-1-reth') bash -c "tar -xzvf chaos-linux-amd64-${version}.tar.gz"
+# 	docker exec $(shell docker ps -aq --filter name='el-1-reth') bash -c "./target/release/chaos --database-path /data/reth/execution-data/db"
+# 	docker cp $(shell docker ps -aq --filter name='el-1-reth'):state.json ../contracts/deployments/state-diff.json
```
ahmadkaouk: We can probably clean this.

test/cli/index.ts
```diff
@@ -144,6 +144,15 @@ program
   .option("--b, --blockscout", "Enable Blockscout")
   .option("--slot-time <number>", "Set slot time in seconds", parseIntValue)
   .option("--cn, --clean-network", "Always clean Kurtosis enclave and Docker containers")
+  .option(
+    "--ic, --inject-contracts",
+    "Inject pre-deployed contracts from state-diff.json into Kurtosis network",
+    true
+  )
```
ahmadkaouk, line 151: The CLI is now configured to inject contracts by default, but if also --`deploy-contracts` is passed, the network would start with preloaded state and then immediately try to redeploy contracts on top of it. There should be a check to skip the deployContracts step if injectContracts is enabled, even if `--deploy-contracts` is passed

test/scripts/check-generated-state.js
undercover-cactus: Like discuss `git checkout` don't create all the files freshly (see https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/Git_FAQ.html#Why_isn.27t_Git_preserving_modification_time_on_files.3F). It will put the time of when you pulled the changes so that building tool can work has expected which also work for us.

gonzamontiel (approved):
Overall looks good, we can add iterative improvements in a follow up PR.
The concerns about the validation logic are outdated. Now, assuming the workflow “generate state → write state-diff.json and state-diff.checksum together” is followed, this PR does effectively guarantee that state-diff.json is up-to-date with the contracts.

contracts/lib/snowbridge
stiiifff, line 1: ⚠️ This is reverting https://github.com/datahaven-xyz/datahaven/pull/361.

.github/workflows/task-e2e.yml
```diff
@@ -48,9 +48,12 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+      - run: ls -lh
```
stiiifff: Is this needed ?

test/contract-bindings/generated.ts
stiiifff, line 1: As mentioned above, this is a revert of https://github.com/datahaven-xyz/datahaven/pull/361.

.github/workflows/task-e2e.yml
undercover-cactus: No it is not needed. It was to check the timestamp on the file because we were using modification time to determine if we needed to generate new state. It was to show that `git checkout` only change the date of only modified files (same as for building system).

However now we don't use this strategy. I removed it.

PR #332, opened 2025-12-03
feat: ✨  enable AVS owner workflow

gonzamontiel:
# Enable AVS owner workflow

Until now, the deployer of the contracts and the owner of the deployed contracts where the same account. Even if we allowed a different owner to be specified, we were using the same. For this reason, a private key was required, so after the deployment we could execute owned transactions needed for the CLI.

In this PR we:
- Add a mechanism to the CLI to specify a different owner account other than the deployer via `--avs-owner-address`
- Add CLI flags `--avs-owner-key` and`--execute-owner-transactions` so account ownership vs. immediate execution is explicit and deferred. If both previous parameters are provided, the CLI will execute the transactions using the private key provided.
- Allow DataHaven AVS deploy scripts to toggle owner-call execution via an env flag `TX_EXECUTION`
- Add documentation on how the new parameters work in `test/README.md` and `test/docs/deployment.md`.

### DIFF
--- a/operator/Cargo.lock
+++ b/operator/Cargo.lock
@@ -8433,7 +8433,7 @@ dependencies = [
 [[package]]
 name = "pallet-bucket-nfts"
 version = "0.1.0"
-source = "git+https://github.com/Moonsong-Labs/storage-hub.git?tag=v0.2.3#527f303055e6ac136db4d6ca373ecc8fd6feeec6"
+source = "git+https://github.com/Moonsong-Labs/storage-hub.git?tag=v0.2.5#24f4489e10159ef08f29fb6b7299e86282f98d79"
 dependencies = [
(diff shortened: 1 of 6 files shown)


test/cli/handlers/contracts/update-metadata.ts
```diff
@@ -73,3 +101,10 @@ export const updateAVSMetadataURI = async (chain: string, uri: string) => {
     throw error;
   }
 };
+
+const normalizePrivateKey = (key?: string): `0x${string}` | undefined => {
+  if (!key) {
+    return undefined;
```
undercover-cactus, line 107: Should be fine to throw an error instead of returning `undefined`. Would avoid issue with having an undefined key used in a script and locking us out.

PR #337, opened 2025-12-03
fix: pallet collective weights aliases

gonzamontiel:
We have two instances of `pallet_collective`
- `pallet_collective_treasury_council`
- `pallet_collective_technical_committee`

Our weights template automatically generates an implementation for `pallet_collective_treasury_council::WeightInfo` or `pallet_collective_technical_committee::WeightInfo`, which don't exist, making the compilation fail right after running benches.
I created aliases for both pallet names, and added a small tweak to the template so it does not break anymore.

### DIFF
--- a/operator/runtime/mainnet/src/lib.rs
+++ b/operator/runtime/mainnet/src/lib.rs
@@ -31,9 +31,11 @@ pub mod weights;
 pub use configs::governance;
 pub use configs::Precompiles;
 
-// Aliases for instanced collective pallets so weights & benchmarks can refer to them
-pub use pallet_collective as pallet_collective_treasury_council;
-pub use pallet_collective as pallet_collective_technical_committee;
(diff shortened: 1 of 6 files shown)


operator/runtime/mainnet/src/lib.rs
```diff
@@ -27,11 +27,14 @@ mod benchmarks;
...
 pub mod precompiles;
 pub mod weights;
-
 // Re-export governance for tests
 pub use configs::governance;
 pub use configs::Precompiles;
 
+// Aliases for instanced collective pallets so weights & benchmarks can refer to them
+pub use pallet_collective as pallet_collective_treasury_council;
+pub use pallet_collective as pallet_collective_technical_committee;
```
stiiifff: I don't understand how that works .. defining two aliases for the same pallet ?

PR #343, opened 2025-12-08
fix: 🛡️  Check origin for validator set messages

gonzamontiel:
### Context
The function `v2_sendMessage()` on Snowbridge Gateway contract is **permissionless** (I'm shocked this is the design choice). Any EOA/contract on Ethereum can build a message and send it through our DH bridge. While we don't change our Snowbridge fork, then this will continue to be the case.

### Problem
We use `v2_sendMessage()` to send **permissioned** operations to our chain. For instance: update our validator set message (coming next, _slashing-related_ messages). So we do need to restrict the processing of the incoming messages on the Substrate side.

### Fix
- I've added a check to `EigenLayerMessageProcessor` that enforces `message.origin` to be only a configured `AuthorisedOrigin`.
- I've added an `AuthorisedOrigin` to `pallet_external_validators::Config`
- I've configured the `AuthorisedOrigin` to be `DatahavenServiceManagerAddress` in all three runtimes

### Stages
- [x] Implementation
- [x] Runtime integration tests
- [x] Collect `DatahavenServiceManagerAddress` parameter for e2e tests to work

Fixes https://github.com/datahaven-xyz/sr-datahaven/issues/12

### DIFF
--- a/test/suites/validator-set-update.test.ts
+++ b/test/suites/validator-set-update.test.ts
@@ -360,10 +360,10 @@ describe("Validator Set Update", () => {
     });
 
     if (!externalValidatorsSetEvent.data) {
-      logger.error("ExternalValidatorsSet event not found");
-      throw new Error("ExternalValidatorsSet event not found");
+      logger.warn("ExternalValidatorsSet event not observed; will rely on storage check.");
+    } else {
+      logger.success("ExternalValidatorsSet event found");
     }
-    logger.success("ExternalValidatorsSet event found");
 
     logger.info(
       "🔍 Checking the new validators are present in the ExternalValidators pallet storage..."


operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1636,7 +1638,9 @@ impl pallet_external_validator_slashes::SendMessage<AccountId> for SlashesSendAd
         let calldata = Vec::new();
 
         let command = Command::CallContract {
-            target: runtime_params::dynamic_params::runtime_config::DatahavenAVSAddress::get(),
+            target:
+                runtime_params::dynamic_params::runtime_config::DatahavenServiceManagerAddress::get(
```
undercover-cactus, line 1642: I am not a huge fan of name that contain `Service`, `Manager`, `Controller` or `Handler`. I find them vague and don't add meaningful information. We could even name it `DatahavenManagerServiceAddress` and that would be the same.
A good practice would maybe to focus on the feature it is associated with. Like it has been done for `AuthorizedOrigin`.

This is just a comment for a discussion. No change needed here.

operator/primitives/bridge/src/lib.rs
```diff
@@ -90,7 +90,7 @@ where
         };
         let decode_result = Self::decode_message(payload.as_slice());
         if let Ok(payload) = decode_result {
-            payload.message_id == EL_MESSAGE_ID
+            payload.message_id == EL_MESSAGE_ID && message.origin == T::AuthorizedOrigin::get()
```
undercover-cactus, line 93: Is Snowbridge setting the origin of the message when relaying them from Ethereum to Substrate ? Or is this like in Ethereum where the origin if the account that is at the origin of the signature ?

operator/runtime/mainnet/src/configs/mod.rs
gonzamontiel: This field refers to the name of an already existing contract, it's not arbitrary. The name of the contract (specially the "ServiceManager" part) follows the pattern of EigenLayer's naming. While it is not mandatory that we name it that way, we should be consistent with the EigenLayer's naming and ecosystem. 
Check https://github.com/Layr-Labs/eigenlayer-middleware/blob/dev/src/ServiceManagerBase.sol, the base contract for all the `ServiceManager` contracts to inherit from.

operator/primitives/bridge/src/lib.rs
gonzamontiel: Yes, exactly. The Gateway contract takes care of that. When a user calls `v2_sendMessage()`, the `msg.sender` goes as the `origin` [here](https://github.com/Moonsong-Labs/snowbridge/blob/solochain/contracts/src/v2/Calls.sol#L60-L62), and then the payload is crafted in [these lines](https://github.com/Moonsong-Labs/snowbridge/blob/solochain/contracts/src/v2/Calls.sol#L111-L119) using that same `origin`.

undercover-cactus: Thank you for answering my question. Sounds good to me. I will approve.

stiiifff: Is there also a way to verify the message signature ? The `who` parameter is unused in both `can_process_message` and `process_message`.

gonzamontiel: The `who` parameter is the caller of `submit()` (see [here](https://github.com/datahaven-xyz/datahaven/blob/fix/check-origin-for-validator-set-messages/operator/pallets/inbound-queue-v2/src/lib.rs#L190-L191)), and this is actually meant to be the *relayer*, who's able to craft a message proving that the an event happened in Ethereum. Actually any Substrate account could pay the fees and provide a valid Ethereum event proof. But it will need to present a valid message. The message relayed is verified in several layers: first the event proof is verified, then we check the message was originated by the Gateway, and since this PR, that the original caller of the message (`message.origin`) was the `DataHavenServiceManager` contract.

PR #345, opened 2025-12-08
feat : Slashing integration in EigenLayer and Datahaven AVS

undercover-cactus:
## Summary

This PR integrate the slashing feature with EigenLayer. With this PR, slashing can now be relayed to our Datahaven AVS and then executed within EigenLayer. In addition some refactoring of the original slashing pallet has been done.

## Motivation 

To avoid misbehaving actor in the network, Datahaven has implemented a slashing pallet in which offenses can be reported and then if adequate can lead to a sanction on the misbehaving node. It incentive nodes to only follow good behavior in addition to the reward incentive. The rewards flow is managed directly into EigenLayer (see https://github.com/datahaven-xyz/datahaven/pull/351).

## Slashing flow

Slashing Flow

## What changes

* Implemented `slashValidatorsOperator` in `DataHavenServiceManager`. It received all the slashing requests batched (every new era the queued slashing are being relayed from substrate to Ethereum). It handle the slashing of the operators reported into the Validator set.
* Added a `slashes_adapter.rs` utility file to remove the duplication for each runtime. In addition, we made use of the `sol!` macro from alloy to encode the calldata for the Ethereum call. This avoid rewriting encoding logic and allow to remove the hardcoded selector value used to call the slashing function.
* Added some tests in solidity to test the registering and slashing of an operator in Ethereum via Eigen Layer.
* Added e2e tests that test the injection of a slash request, it being relayed via the snowbridge relayer and executed by our Datahaven AVS.

## What could be better

* We are only deploying one strategy for now so it is hardcoded in the slashing flow. We should be able to update the pallet in case we are adding a new strategy. So communication from Ethereum should be relayed.
* We don't have error being return in case the slashing fail. Which could happen if we don't have the right number of strategy or the validator is not registered... etc.
* More tests for the unhappy path

### DIFF
--- a/operator/runtime/common/src/slashes_adapter.rs
+++ b/operator/runtime/common/src/slashes_adapter.rs
@@ -45,7 +45,7 @@ pub trait SlashesSubmissionConfig {
     fn slashes_agent_origin() -> H256;
 
     /// Get the strategies to slash.
-    fn strategies() -> Vec<H160>;
+    fn strategies() -> Vec<Address>;
 }
 
 /// Generic slashes submission adapter.
@@ -98,24 +98,19 @@ impl<C: SlashesSubmissionConfig> pallet_external_validator_slashes::SendMessage<
 
 fn encode_slashing_request(
     slashes_utils: &Vec<SlashData<AccountId>>,
-    strategies: Vec<H160>,
+    strategies: Vec<Address>,
 ) -> Vec<u8> {
     let mut slashings: Vec<SlashingRequest> = vec![];
+    let strategies_len = strategies.len();
 
     // Extend with operator address to slash
     for slash_operator in slashes_utils {
         // slashing all the strategies
-        let wads_to_slash = strategies
-            .iter()
-            .map(|_| U256::from(slash_operator.wad_to_slash))
-            .collect();
+        let wads_to_slash = vec![U256::from(slash_operator.wad_to_slash); strategies_len];
 
         let slashing_request = SlashingRequest {
             operator: Address::from(slash_operator.validator.0),
-            strategies: strategies
-                .iter()
-                .map(|s| Address::from(s.as_fixed_bytes()))
-                .collect(),
+            strategies: strategies.clone(),
             wadsToSlash: wads_to_slash, // We only have one strategy deployed
             description: "Slashing validator".into(),
         };
--- a/operator/runtime/mainnet/src/configs/mod.rs
+++ b/operator/runtime/mainnet/src/configs/mod.rs
@@ -30,6 +30,7 @@ use super::{
     Signature, System, Timestamp, Treasury, TxPause, BLOCK_HASH_COUNT, EXTRINSIC_BASE_WEIGHT,
     MAXIMUM_BLOCK_WEIGHT, NORMAL_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO, SLOT_DURATION, VERSION,
 };
+use alloy_core::primitives::Address;
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_runtime::{traits::AccountIdConversion, RuntimeDebug};
@@ -1693,12 +1694,12 @@ impl datahaven_runtime_common::slashes_adapter::SlashesSubmissionConfig for Main
         // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
     }
 
-    fn strategies() -> Vec<H160> {
+    fn strategies() -> Vec<Address> {
         // We only slash strategy that we reward
-        let mut strategies: Vec<H160> =
+        let mut strategies: Vec<Address> =
             runtime_params::dynamic_params::runtime_config::RewardsStrategiesAndMultipliers::get()
                 .iter()
-                .map(|(strategy, _mult)| strategy.clone())
+                .map(|(strategy, _mult)| Address::from(strategy.as_fixed_bytes()))
                 .collect();
         // The array of strategies need to be in ascending order (see https://github.com/Layr-Labs/eigenlayer-contracts/blob/7ecc83c7b180850531bc5b8b953a7340adeecd43/src/contracts/core/AllocationManager.sol#L343-L347)
         strategies.sort();
--- a/operator/runtime/stagenet/src/configs/mod.rs
+++ b/operator/runtime/stagenet/src/configs/mod.rs
@@ -30,6 +30,7 @@ use super::{
     Signature, System, Timestamp, Treasury, TxPause, BLOCK_HASH_COUNT, EXTRINSIC_BASE_WEIGHT,
     MAXIMUM_BLOCK_WEIGHT, NORMAL_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO, SLOT_DURATION, VERSION,
 };
+use alloy_core::primitives::Address;
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_runtime::{traits::AccountIdConversion, RuntimeDebug};
@@ -1689,12 +1690,12 @@ impl datahaven_runtime_common::slashes_adapter::SlashesSubmissionConfig for Stag
         // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
     }
 
-    fn strategies() -> Vec<H160> {
+    fn strategies() -> Vec<Address> {
         // We only slash strategy that we reward
-        let mut strategies: Vec<H160> =
+        let mut strategies: Vec<Address> =
             runtime_params::dynamic_params::runtime_config::RewardsStrategiesAndMultipliers::get()
                 .iter()
-                .map(|(strategy, _mult)| strategy.clone())
+                .map(|(strategy, _mult)| Address::from(strategy.as_fixed_bytes()))
                 .collect();
         // The array of strategies need to be in ascending order (see https://github.com/Layr-Labs/eigenlayer-contracts/blob/7ecc83c7b180850531bc5b8b953a7340adeecd43/src/contracts/core/AllocationManager.sol#L343-L347)
         strategies.sort();
--- a/operator/Cargo.lock
+++ b/operator/Cargo.lock
@@ -2609,6 +2609,7 @@ dependencies = [
 name = "datahaven-mainnet-runtime"
 version = "0.13.0"
 dependencies = [
+ "alloy-core",
  "bridge-hub-common 0.13.1",
  "datahaven-runtime-common",
  "dhp-bridge",
@@ -3059,6 +3060,7 @@ dependencies = [
 name = "datahaven-testnet-runtime"
 version = "0.13.0"
 dependencies = [
+ "alloy-core",
  "bridge-hub-common 0.13.1",
  "datahaven-runtime-common",
  "dhp-bridge",
--- a/operator/pallets/external-validator-slashes/src/benchmarking.rs
+++ b/operator/pallets/external-validator-slashes/src/benchmarking.rs
@@ -102,7 +102,7 @@ mod benchmarks {
 
         #[block]
         {
-            processed = Pallet::<T>::process_slashes_queue(s);
+            processed = Pallet::<T>::process_slashes_queue(s).unwrap();
         }
 
         assert_eq!(UnreportedSlashesQueue::<T>::get().len(), 1);
--- a/operator/runtime/mainnet/Cargo.toml
+++ b/operator/runtime/mainnet/Cargo.toml
@@ -13,6 +13,7 @@ version = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
+alloy-core = { workspace = true, features = ["sol-types"] }
 bridge-hub-common = { workspace = true, optional = true }
 codec = { workspace = true, features = ["derive"] }
 datahaven-runtime-common = { workspace = true }
--- a/operator/runtime/testnet/Cargo.toml
+++ b/operator/runtime/testnet/Cargo.toml
@@ -13,6 +13,7 @@ version = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
+alloy-core = { workspace = true, features = ["sol-types"] }
 bridge-hub-common = { workspace = true, optional = true }
 codec = { workspace = true, features = ["derive"] }
 datahaven-runtime-common = { workspace = true }
@@ -183,6 +184,7 @@ snowbridge-pallet-system-v2 = { workspace = true }
 [features]
 default = ["std"]
 std = [
+    "alloy-core/std",
     "codec/std",
     "datahaven-runtime-common/std",
     "fp-account/std",
--- a/operator/runtime/testnet/src/configs/mod.rs
+++ b/operator/runtime/testnet/src/configs/mod.rs
@@ -30,6 +30,7 @@ use super::{
     Signature, System, Timestamp, Treasury, TxPause, BLOCK_HASH_COUNT, EXTRINSIC_BASE_WEIGHT,
     MAXIMUM_BLOCK_WEIGHT, NORMAL_BLOCK_WEIGHT, NORMAL_DISPATCH_RATIO, SLOT_DURATION, VERSION,
 };
+use alloy_core::primitives::Address;
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_runtime::{traits::AccountIdConversion, RuntimeDebug};
@@ -1693,12 +1694,12 @@ impl datahaven_runtime_common::slashes_adapter::SlashesSubmissionConfig for Test
         // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
     }
 
-    fn strategies() -> Vec<H160> {
+    fn strategies() -> Vec<Address> {
         // We only slash strategy that we reward
-        let mut strategies: Vec<H160> =
+        let mut strategies: Vec<Address> =
             runtime_params::dynamic_params::runtime_config::RewardsStrategiesAndMultipliers::get()
                 .iter()
-                .map(|(strategy, _mult)| strategy.clone())
+                .map(|(strategy, _mult)| Address::from(strategy.as_fixed_bytes()))
                 .collect();
         // The array of strategies need to be in ascending order (see https://github.com/Layr-Labs/eigenlayer-contracts/blob/7ecc83c7b180850531bc5b8b953a7340adeecd43/src/contracts/core/AllocationManager.sol#L343-L347)
         strategies.sort();


operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -579,7 +592,7 @@ impl<T: Config> Pallet<T> {
 
                 slashes_to_send.push(SlashData {
                     validator: slash.validator,
-                    amount_to_slash: 0, // TODO: need to compute how much we slash
+                    amount_to_slash: u128::from_str_radix("10000000000000000", 10).unwrap(), // TODO: need to compute how much we slash (for now it is 1e16)
```
stiiifff: This shouldn't be a hard-coded amount. Let's discuss about what would be the right way to approach this.

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1678,51 +1678,26 @@ impl pallet_tx_pause::Config for Runtime {
...
+
+    fn service_manager_address() -> H160 {
+        runtime_params::dynamic_params::runtime_config::DatahavenServiceManagerAddress::get()
     }
-    fn deliver(message: Self::Ticket) -> Result<H256, SendError> {
-        EthereumOutboundQueueV2::deliver(message)
+
+    fn slashes_agent_origin() -> H256 {
+        runtime_params::dynamic_params::runtime_config::RewardsAgentOrigin::get()
+        // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
```
stiiifff, line 1694: We should consider separate origins for rewards & slashing (i.e. security isolation).

operator/runtime/stagenet/src/configs/mod.rs
```diff
@@ -1674,51 +1674,26 @@ impl pallet_tx_pause::Config for Runtime {
...
+
+    fn service_manager_address() -> H160 {
+        runtime_params::dynamic_params::runtime_config::DatahavenServiceManagerAddress::get()
     }
-    fn deliver(message: Self::Ticket) -> Result<H256, SendError> {
-        EthereumOutboundQueueV2::deliver(message)
+
+    fn slashes_agent_origin() -> H256 {
+        runtime_params::dynamic_params::runtime_config::RewardsAgentOrigin::get()
+        // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
```
stiiifff, line 1690: Same remark as previously.

operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -327,19 +327,14 @@ pub mod dynamic_params {
...
-        pub static SlashOperatorSelector: BoundedVec<u8, ConstU32<4>> =
-            BoundedVec::truncate_from(vec![0xca, 0x48, 0x11, 0x9f]);
-
-        #[codec(index = 36)]
-        #[allow(non_upper_case_globals)]
         /// The Ethereum address of the DataHavenServiceManager contract.
         /// This address is used both for authorized slashing requests and validator-set update messages.
-        pub static DatahavenServiceManagerAddress: H160 = H160::repeat_byte(0x0);
+        pub static DatahavenServiceManagerAddress: H160 =
+            H160::from_slice(&hex!("809d550fca64d94Bd9F66E60752A544199cfAC3D"));
```
stiiifff: Which address is that ?

operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -327,19 +327,14 @@ pub mod dynamic_params {
...
 
         // ╔══════════════════════ Validator Rewards Inflation ═══════════════════════╗
 
-        #[codec(index = 37)]
+        #[codec(index = 36)]
```
stiiifff: I would not change the runtime dynamic parameter indexes anymore at this time. Just remove `SlashOperatorSelector`, and add a placeholder comment.

operator/pallets/external-validator-slashes/src/lib.rs
stiiifff, line 1: There's multiple occurrences of a wrong `ext_validators_rewards` log target.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -351,6 +359,34 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
         _allocationManager.createOperatorSets(address(this), operatorSets);
     }
 
+    // ============ Slashing Submitter Functions ============
+
+    /**
+     * @notice Slash the operators of the validators set
+     * @dev This function should be called during initialisation to set up the required operator set.
+     */
```
ahmadkaouk:
```suggestion
```
this comment is not correct

operator/runtime/mainnet/src/configs/mod.rs
undercover-cactus: I don't think there is advantage in separating them. In the contrary we could set one and forget the other. It add complexity and therefore a security risk. We can discuss it.

operator/pallets/external-validator-slashes/src/lib.rs
undercover-cactus: Fixed. I replaced them with `ext_validators_slashes`

operator/pallets/external-validator-slashes/src/lib.rs
undercover-cactus: Offenses have already a percent defined for each offense. Would it be ok to use those ?

operator/runtime/stagenet/src/configs/runtime_params.rs
undercover-cactus: This is the address of the proxy for the Datahaven AVS. I notice we remove it so it is an error from rebasing. I am reverting this change.

operator/runtime/stagenet/src/configs/runtime_params.rs
undercover-cactus: I have reverted the index change.

operator/pallets/external-validator-slashes/src/lib.rs
stiiifff: I think so, yes. The formula to calculate the slashing percentage (to be applied on validators' stake) is well documented in the Polkadot docs, and I think applying the same formula and slashing percentages make sense (on operators' stakes allocated to the different strategies). So it seems reasonable to relay the slash percentage and reasons (via Snowbridge) to the EL slashing contract.

operator/runtime/mainnet/src/configs/mod.rs
stiiifff: I'm ok with re-using the RewardsAgentOrigin for now.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -60,6 +60,14 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
     IGatewayV2 private _snowbridgeGateway;
 
+    struct SlashingRequest {
+        address operator;
+        uint256[] wadsToSlash;
+        string description;
+    }
+
+    event SlashingComplete();
+
```
ahmadkaouk: This should be moved to IDatahavenServiceManager.sol for consistency

contracts/src/DataHavenServiceManager.sol
```diff
@@ -351,6 +359,34 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
+     * @notice Slash the operators of the validators set
+     * @dev This function should be called during initialisation to set up the required operator set.
+     */
+    function slashValidatorsOperator(
+        SlashingRequest[] calldata slashings
+    ) external onlyRewardsInitiator {
+        OperatorSet memory operatorSet = OperatorSet({avs: address(this), id: VALIDATORS_SET_ID});
+        IStrategy[] memory strategies = _allocationManager.getStrategiesInOperatorSet(operatorSet);
+
+        for (uint256 i = 0; i < slashings.length; i++) {
```
ahmadkaouk, line 365: I would prevalidate inputs and add this check `strategies.length == wadsToSlash.length`

contracts/src/DataHavenServiceManager.sol
```diff
@@ -351,6 +359,34 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
+            IAllocationManagerTypes.SlashingParams memory slashingParams =
+                IAllocationManagerTypes.SlashingParams({
+                    operator: slashings[i].operator,
+                    operatorSetId: VALIDATORS_SET_ID,
+                    strategies: strategies,
+                    wadsToSlash: slashings[i].wadsToSlash,
+                    description: slashings[i].description
+                });
```
ahmadkaouk, line 373: I believe that passing all the strategies of an operator set, as we do here in `slashingParams` is a fragile approach adn error prone:
- If we ever add/remove a strategy from the operator set (runtime must immediately adapt  for wads so a runtime upgrade might be needed), 
- Or If we want to slash only a subset of strategies than this wont work

I think a better approach here is to specify the strategies in the slashing request that comes from the runtime.

contracts/test/Slashing.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        vm.prank(avsOwner);
+        serviceManager.addValidatorToAllowlist(operator);
+
+        // Configure the rewards initiator (because only the reward agent can submit slashing request)
+        vm.prank(avsOwner);
+        serviceManager.setRewardsInitiator(snowbridgeAgent);
+
+        vm.prank(operator);
+        delegationManager.registerAsOperator(address(0), 0, "");
+
```
ahmadkaouk: We are missing the step of allocating magnitude/delegate stake to the operator. I believe in this case the magnitude is always zero and in this case EigenLayer _slashOperator hit if (allocation.currentMagnitude == 0) continue; and no slashing is happening.

contracts/test/Slashing.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        uint256[] memory wadsToSlash = new uint256[](3); // 3 wadsToSlash because we have register 3 strategies for the Validator set
+        wadsToSlash[0] = 1e16;
+        wadsToSlash[1] = 1e16;
+        wadsToSlash[2] = 1e16;
+
+        slashings[0] =
+            DataHavenServiceManager.SlashingRequest(operator, wadsToSlash, "Testing slashing");
+
+        console.log(block.number);
+        vm.roll(block.number + uint32(7 days) + 1);
```
ahmadkaouk, line 68: I am confused here; why are we adding a time duration(7 days) to a block number?

contracts/test/Slashing.t.sol
```diff
@@ -0,0 +1,77 @@
...
+contract SlashingTest is AVSDeployer {
+    address operator = address(0xabcd);
+    address public snowbridgeAgent = address(uint160(uint256(keccak256("snowbridgeAgent"))));
+
+    function setUp() public virtual {
+        _deployMockEigenLayerAndAVS();
+    }
+
+    function test_fulfilSlashingRequest() public {
+        // Allow our operatir to register
```
ahmadkaouk:
```suggestion
        // Allow our operator to register
```
typo

contracts/test/Slashing.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        console.log(block.number);
+        vm.roll(block.number + uint32(7 days) + 1);
+        console.log(block.number);
+        vm.expectEmit();
+
+        OperatorSet memory operatorSet =
+            OperatorSet({avs: address(serviceManager), id: serviceManager.VALIDATORS_SET_ID()});
+        IStrategy[] memory strategies = allocationManager.getStrategiesInOperatorSet(operatorSet);
+
+        // Because the current magnituse for the allocation is 0
```
ahmadkaouk:
```suggestion
        // Because the current magnitude for the allocation is 0
```
typo

operator/pallets/external-validator-slashes/src/lib.rs
ahmadkaouk: We should never use `unwrap()` in the runtime. This will panic on any error and will break the chain.

ahmadkaouk: We should never use `unwrap()` in the runtime. This will panic on any error and will break the chain.

operator/runtime/common/src/slashes_adapter.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+fn encode_slashing_request(slashes_utils: &Vec<SlashData<AccountId>>) -> Vec<u8> {
+    let mut slashings: Vec<SlashingRequest> = vec![];
+
+    // Extend with operator address to slash
+    for slash_operator in slashes_utils.into_iter() {
+        let slashing_request = SlashingRequest {
+            operator: Address::from(slash_operator.validator.0),
+            wadsToSlash: vec![U256::from(slash_operator.amount_to_slash)], // We only have one strategy deployed
```
ahmadkaouk: Instead of hardcoding the strategies here, it would be better to have a runtime parameter of strategies used for slashing

ahmadkaouk: Also the variable name `amount_to_slash` is misleading since it strongly suggests an absolute amount while it is representing a fraction represented as fixed point with 18 decimals.

test/suites/slash.test.ts
ahmadkaouk, line 1: There seems to be no test for the Ethereum side, such as verifying that the slashOperator function in DataHavenServiceManager was successfully called.

test/suites/slash.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      percentage: 20
+    });
+    const sudoTx = dhApi.tx.Sudo.sudo({
+      call: sudoSlashCall.decodedCall
+    });
+    const alithSigner = getPapiSigner("ALITH");
+    const resultSubmitTx = await sudoTx.signAndSubmit(alithSigner);
+    expect(resultSubmitTx.ok).toBeTruthy();
+
+    console.log("Transaction submitted !");
```
ahmadkaouk: Can we replace `console.log` with `logger.debug` to maintain consistency with other tests and simplify the process of enabling or disabling those logs?

contracts/src/DataHavenServiceManager.sol
undercover-cactus: I removed it and added a description of the parameters.

contracts/src/DataHavenServiceManager.sol
undercover-cactus: I have moved both the struct and the event to IDatahavenServiceManager.sol for consistency.

contracts/src/DataHavenServiceManager.sol
undercover-cactus: This check is already being done in the `slashOperator` function. It would be a duplicate.

See https://github.com/Layr-Labs/eigenlayer-contracts/blob/main/src/contracts/core/AllocationManager.sol#L61C1-L67C100

contracts/src/DataHavenServiceManager.sol
undercover-cactus: If we want to slash a subset we can just put 0 for the wadToSlash for this startegy.

For the first point, I agree it is not the best. I am doing the change but then if we remove a strategy we need to first remove it in the runtime to avoid any error.

ahmadkaouk: Actually, the allocation manager doesn’t allow `wadToSlash` to be zero. If this is the case, it will revert. https://github.com/Layr-Labs/eigenlayer-contracts/blob/80b74fa3b9908190fa6d8396778d1896a4bfb4dc/src/contracts/core/AllocationManager.sol#L427-L428

undercover-cactus: I have done the changes. To keep things simple I am using the strategies defined in `RewardsStrategiesAndMultipliers`. So we are assuming that we want to slash the same strategies we are rewarding.

contracts/test/Slashing.t.sol
undercover-cactus: The slashing is happening but yes the wads actually slash for the strategy is 0. I still have the event being emitted from EL.

contracts/test/Slashing.t.sol
undercover-cactus: Yeah I find it a bit odd too. It was to match the value in the AVSDeployer (see https://github.com/datahaven-xyz/datahaven/blob/main/contracts/test/utils/AVSDeployer.sol#L169).

Maybe here we should have used blocks instead of the time units.

operator/pallets/external-validator-slashes/src/lib.rs
undercover-cactus: fixed

undercover-cactus: fixed

operator/runtime/common/src/slashes_adapter.rs
undercover-cactus: I changed the name to `wad_to_slash` because indeed it is not the amount bu a proportion of the share.

I have changed the code to use the strategies defined in the runtime for the reward. So I am assuming we want to slash the strategies we are rewarding (it avoid having to update 2 runtime variables).

contracts/test/Slashing.t.sol
ahmadkaouk: Yes I agree we should use blocks instead of time units in AVSDeployer, it's a bug.

operator/runtime/common/src/slashes_adapter.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn deliver(message: Self::Ticket) -> Result<H256, SendError> {
+        C::OutboundQueue::deliver(message)
+    }
+}
+
+fn encode_slashing_request(slashes_utils: &Vec<SlashData<AccountId>>, strategies: Vec<H160>) -> Vec<u8> {
+    let mut slashings: Vec<SlashingRequest> = vec![];
+
+    // Extend with operator address to slash
+    for slash_operator in slashes_utils.into_iter() {
```
ahmadkaouk:
```suggestion
    for slash_operator in slashes_utils {
```
nit

operator/runtime/common/src/slashes_adapter.rs
```diff
@@ -0,0 +1,121 @@
...
+        // slashing all the strategies
+        let wads_to_slash = strategies.iter().map(|_| U256::from(slash_operator.wad_to_slash)).collect();
+
+        let slashing_request = SlashingRequest {
+            operator: Address::from(slash_operator.validator.0),
+            strategies: strategies.iter().map(|s| Address::from(s.as_fixed_bytes())).collect(),
```
ahmadkaouk: the `AllocationManager` require strategies to be in ascending order so we should make sure they are sorted.
https://github.com/Layr-Labs/eigenlayer-contracts/blob/7ecc83c7b180850531bc5b8b953a7340adeecd43/src/contracts/core/AllocationManager.sol#L343-L347

ahmadkaouk: Also we are converting `strategies` to `Vec<Address>` on every iteration, but the result is the same each time. It can be computed once before the loop.

operator/runtime/common/src/slashes_adapter.rs
undercover-cactus: Fixed

operator/runtime/common/src/slashes_adapter.rs
undercover-cactus: I added a call to `.sort` to get the strategies in the right order.

operator/pallets/external-validator-slashes/src/lib.rs
gonzamontiel: I was also concerned about this one, is it going to be fixed in this PR?

contracts/src/DataHavenServiceManager.sol
```diff
@@ -351,6 +351,33 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
         _allocationManager.createOperatorSets(address(this), operatorSets);
     }
 
+    // ============ Slashing Submitter Functions ============
+
+    /**
+     * @notice Slash the operators of the validators set
+     * @param slashings array of request to slash operator containing the operator to slash, array of proportions to slash and the reason of the slashing.
+     */
+    function slashValidatorsOperator(
```
gonzamontiel, line 360: nit: I would name it something like:
```suggestion
    function executeSlashingRequests(
```

ahmadkaouk (approved):

PR #346, opened 2025-12-08
feat: add feature to build binary with postgres bundled

undercover-cactus:
## Summary

Re-add the static build feature option to bundle postgres dependency into the binary. It simplify the installation because now to run the node the operator doesn't need to have postgres dependencies installed on its system.

## What changed ?

* Added a `static` feature that can be activated to add the extra dependencies during the build.
* A task that run every time a dependency has been modified so we can make sure the build with the feature is still working correctly. (we are assuming simple change in the code won't have an impact on it because postgres is being used in diesel which is not a direct dependecy to datahaven).


stiiifff (approved):
LGTM 👍

PR #351, opened 2025-12-10
feat: Implement EigenLayer Rewards V2 distribution

ahmadkaouk:
### Summary

This PR implements the EigenLayer Rewards Distribution V2 model for DataHaven, replacing the previous merkle-root-based rewards registry approach with EigenLayer's native `OperatorDirectedRewardsSubmission` API. This enables direct integration with EigenLayer's RewardsCoordinator for validator rewards distribution.

### Motivation

EigenLayer's V2 rewards model provides several advantages:
- **Direct integration**: Uses EigenLayer's native `createOperatorDirectedOperatorSetRewardsSubmission` API
- **Per-operator rewards**: Distributes rewards proportionally to individual operators based on their earned points
- **Simplified architecture**: Removes the need for a separate RewardsRegistry contract
- **Better UX**: Operators receive rewards directly through EigenLayer's established claiming mechanism

### Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│                       DataHaven Substrate                       │
├─────────────────────────────────────────────────────────────────┤
│  Era End                                                        │
│    │                                                            │
│    ▼                                                            │
│  external-validators-rewards pallet                             │
│    │ generate_era_rewards_utils()                               │
│    │   • Calculate individual points per validator              │
│    │   • Compute total inflation amount                         │
│    │                                                            │
│    ▼                                                            │
│  RewardsSubmissionAdapter (runtime_common)                      │
│    │ build() → points_to_rewards() → encode_rewards_calldata()  │
│    │                                                            │
│    ▼                                                            │
│  Snowbridge Outbound Queue                                      │
│    │ CallContract(ServiceManager.submitRewards(...))            │
└────│────────────────────────────────────────────────────────────┘
     │
     ▼ Cross-chain message via Snowbridge
┌─────────────────────────────────────────────────────────────────┐
│                         Ethereum                                │
├─────────────────────────────────────────────────────────────────┤
│  DataHavenServiceManager                                        │
│    │ submitRewards(OperatorDirectedRewardsSubmission)           │
│    │   • Approve wHAVE tokens to RewardsCoordinator             │
│    │                                                            │
│    ▼                                                            │
│  EigenLayer RewardsCoordinator                                  │
│    │ createOperatorDirectedOperatorSetRewardsSubmission()       │
│    │                                                            │
│    ▼                                                            │
│  Operators claim rewards via EigenLayer                         │
└─────────────────────────────────────────────────────────────────┘
```

#### Smart Contracts (`contracts/`)

**DataHavenServiceManager.sol**
- Added `submitRewards(OperatorDirectedRewardsSubmission)` function to submit rewards to EigenLayer's RewardsCoordinator
- Implements `SafeERC20` for secure token approvals
- Uses `onlyRewardsInitiator` modifier for access control (Snowbridge Agent)
- Emits `RewardsSubmitted` and `RewardsInitiatorSet` events for tracking

**IDataHavenServiceManager.sol**
- Added `submitRewards()` interface for EigenLayer rewards submission
- Added `setRewardsInitiator()` interface for configuring the authorized caller
- Added new events: `RewardsSubmitted`, `RewardsInitiatorSet`

**New Test: RewardsSubmitter.t.sol**
- Comprehensive test suite covering:
  - Access control (only rewards initiator can submit)
  - Single and multiple operator rewards
  - Multiple consecutive submissions
  - Custom descriptions and different tokens

#### Substrate Runtime (`operator/`)

**New: `runtime/common/src/rewards_adapter.rs` (934 lines)**

A generic, configurable adapter for building EigenLayer rewards messages:

- **`RewardsSubmissionConfig` trait**: Runtime-agnostic configuration interface
  - `OutboundQueue`: Snowbridge outbound queue type
  - `rewards_duration()`: Reward period duration (typically 86400s)
  - `whave_token_address()`: wHAVE ERC20 token on Ethereum
  - `service_manager_address()`: ServiceManager contract address
  - `rewards_agent_origin()`: Snowbridge agent origin

- **`RewardsSubmissionAdapter<C>`**: Generic implementation of `SendMessage` trait

- **`points_to_rewards()`**: Converts validator points to token amounts
  - Proportional distribution based on total points
  - Returns remainder (dust) from integer division
  - Arithmetic overflow/underflow protection

- **`encode_rewards_calldata()`**: ABI-encodes the `submitRewards` call
  - Uses `alloy-core` for type-safe Solidity ABI encoding
  - Validates `uint96` multiplier bounds

- **Comprehensive test suite** covering:
  - Basic and edge-case reward calculations
  - Remainder/dust handling
  - Overflow/underflow protection
  - ABI encoding round-trip verification
  - Message building with various configurations

**Modified: `pallets/external-validators-rewards/`**

- **`types.rs`**: Extended `EraRewardsUtils` struct:
  ```rust
  pub struct EraRewardsUtils {
      pub era_index: u32,                    // NEW
      pub rewards_merkle_root: H256,
      pub leaves: Vec<H256>,
      pub leaf_index: Option<u64>,
      pub total_points: u128,
      pub individual_points: Vec<(H160, u32)>, // NEW
      pub inflation_amount: u128,             // NEW
      pub era_start_timestamp: u32       // NEW
  }
  ```

- **`lib.rs`**: Updated `generate_era_rewards_utils()`:
  - Now accepts `inflation_amount` parameter
  - Extracts `individual_points` as `(H160, u32)` tuples for EigenLayer
  - Returns `None` when `total_points` is zero (prevents inflation with no distribution)

- **`mock.rs`**: Updated test mock to use `H160` as `AccountId` (matching DataHaven's EVM-compatible account model)

**Modified: Runtime Configurations**

All three runtimes (mainnet, stagenet, testnet) updated:

1. **New runtime parameters** (`runtime_params.rs`):
   - `ServiceManagerAddress`: DataHaven ServiceManager contract on Ethereum
   - `WHAVETokenAddress`: wHAVE ERC20 token address
   - `RewardsGenesisTimestamp`: EigenLayer-aligned genesis timestamp
   - `RewardsDuration`: Rewards period (default: 86400 = 1 day)

2. **Refactored `RewardsSendAdapter`**:
   - Replaced inline implementation with `RewardsSubmissionAdapter<Config>`
   - Each runtime implements `RewardsSubmissionConfig` trait
   - Cleaner, DRY configuration

## ⚠️ Breaking Changes ⚠️

- **Runtime Parameters**: New parameters must be configured via governance before rewards submission will work:
  - `ServiceManagerAddress` (replaces `RewardsRegistryAddress`)
  - `WHAVETokenAddress`
  - `RewardsGenesisTimestamp`
  
- **Contract Interface**: `submitRewards()` now accepts a full `OperatorDirectedRewardsSubmission` struct instead of a merkle root

### DIFF
--- a/.github/workflows/task-build-static-operator.yaml
+++ b/.github/workflows/task-build-static-operator.yaml
@@ -1,0 +1,45 @@
+name: DataHaven Operator Static Build
+
+on:
+  pull_request:
+    paths:
+      # Only check the build if we are adding a dependency to save runner time
+      - 'operator/Cargo.toml'
+      - 'operator/Cargo.lock'
+
+jobs:
+  build-node:
+    name: Build operator binary
+    runs-on:
+      group: DH-runners
+    env:
+      RUSTC_WRAPPER: "sccache"
+      CARGO_INCREMENTAL: "0"
+      CARGO_TERM_COLOR: always
+      SCCACHE_GHA_ENABLED: "true"
+    defaults:
+      run:
+        working-directory: ./operator
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 1
+
+      - uses: ./.github/workflows/actions/setup-env
+        with:
+          cache-key: BUILD-RELEASE
+          install-deps: false
+          skip-libpq: true
+
+      - name: Set build flags
+        run: echo "RUSTFLAGS=-C linker=clang -C link-arg=-fuse-ld=mold" >> $GITHUB_ENV
+
+      - name: Build node binary
+        run: |
+          cargo build --release --locked --features fast-runtime,static
+
+      - name: Test binary
+        run: |
+          ldd ./target/release/datahaven-node
+          ./target/release/datahaven-node --version
--- a/operator/Cargo.lock
+++ b/operator/Cargo.lock
@@ -2726,6 +2726,7 @@ dependencies = [
  "pallet-transaction-payment-rpc",
  "pallet-transaction-payment-rpc-runtime-api",
  "parity-scale-codec",
+ "pq-sys",
  "sc-basic-authorship",
  "sc-cli",
  "sc-client-api",
@@ -10554,6 +10555,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "pq-src"
+version = "0.3.10+libpq-18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56ef39ce621f4993d6084fdcd4cbf1e01c84bdba53109cfad095d2cf441b85b9"
+dependencies = [
+ "cc",
+ "openssl-sys",
+]
+
+[[package]]
 name = "pq-sys"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10561,6 +10572,7 @@ checksum = "089d5dc8f44104b719912ad4478fd558b59a431ce19ef9101f637be8c656b90a"
 dependencies = [
  "libc",
  "pkg-config",
+ "pq-src",
  "vcpkg",
 ]
 
--- a/operator/Cargo.toml
+++ b/operator/Cargo.toml
@@ -303,6 +303,12 @@ shp-types = { git = "https://github.com/Moonsong-Labs/storage-hub.git", tag = "v
 ## Precompiles
 pallet-evm-precompile-file-system = { git = "https://github.com/Moonsong-Labs/storage-hub.git", tag = "v0.2.9", default-features = false }
 
+
+# Static linking
+#### Needed to build static binaries ####
+pq-sys = { version = "0.7.4" }
+
+
 # The list of dependencies below (which can be both direct and indirect dependencies) are crates
 # that are suspected to be CPU-intensive, and that are unlikely to require debugging (as some of
 # their debug info might be missing) or to require to be frequently recompiled. We compile these
--- a/operator/node/Cargo.toml
+++ b/operator/node/Cargo.toml
@@ -131,6 +131,11 @@ sp-keystore = { workspace = true }
 serde = { workspace = true, default-features = true }
 toml = { workspace = true }
 
+# Static linking
+#### Needed to build static binaries ####
+pq-sys = { workspace = true, optional = true }
+
+
 [build-dependencies]
 substrate-build-script-utils = { workspace = true, default-features = true }
 
@@ -172,3 +177,6 @@ metadata-hash = [
 	"datahaven-mainnet-runtime/metadata-hash",
 	"datahaven-testnet-runtime/metadata-hash",
 ]
+
+# Bundle the postgres source in the binary to avoid having to install the deps in the system (simplify installation)
+static = ["pq-sys", "pq-sys/bundled"]
--- a/operator/scripts/verify-licenses.sh
+++ b/operator/scripts/verify-licenses.sh
@@ -52,6 +52,7 @@ AUTHORS=(
 NAMES=(
     "ring"  # v0.16.20 has null license metadata but contains Apache-2.0 AND ISC LICENSE file
     "shp-tx-implicits-runtime-api"
+    "pq-src" # License is the same as postgres
 )
 licenses_filter=$(printf ' .license != "%s" and' "${LICENSES[@]}")
 authors_filter=$(printf ' .authors != "%s" and' "${AUTHORS[@]}")


operator/runtime/mainnet/src/configs/runtime_params.rs
```diff
@@ -358,6 +358,28 @@ pub mod dynamic_params {
...
+        /// The wHAVE ERC20 token address on Ethereum.
+        /// Used in the OperatorDirectedRewardsSubmission struct.
+        pub static WHAVETokenAddress: H160 = H160::repeat_byte(0x0);
+
+        #[codec(index = 41)]
+        #[allow(non_upper_case_globals)]
+        /// EigenLayer-aligned genesis timestamp for rewards calculation.
+        /// Must be divisible by 86400 (seconds per day) as per EigenLayer requirements.
+        /// Default: 0 (must be set via governance to actual deployment timestamp).
+        pub static RewardsGenesisTimestamp: u32 = 0;
```
gonzamontiel, line 406: I believe this became obsolete since this commit https://github.com/datahaven-xyz/datahaven/pull/351/changes/b8c8ea30ef9a799632cda2bb002bc1b9c37a322f, when the calculation for `era_start_timestamp()` was moved inside the pallet's `on_era_end()`

operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -363,6 +363,28 @@ pub mod dynamic_params {
...
+        /// The wHAVE ERC20 token address on Ethereum.
+        /// Used in the OperatorDirectedRewardsSubmission struct.
+        pub static WHAVETokenAddress: H160 = H160::repeat_byte(0x0);
+
+        #[codec(index = 41)]
+        #[allow(non_upper_case_globals)]
+        /// EigenLayer-aligned genesis timestamp for rewards calculation.
+        /// Must be divisible by 86400 (seconds per day) as per EigenLayer requirements.
+        /// Default: 0 (must be set via governance to actual deployment timestamp).
+        pub static RewardsGenesisTimestamp: u32 = 0;
```
gonzamontiel, line 409: Same as prev comment

operator/runtime/testnet/src/configs/runtime_params.rs
```diff
@@ -359,6 +359,25 @@ pub mod dynamic_params {
...
+
+        #[codec(index = 40)]
+        #[allow(non_upper_case_globals)]
+        /// The wHAVE ERC20 token address on Ethereum.
+        pub static WHAVETokenAddress: H160 = H160::repeat_byte(0x0);
+
+        #[codec(index = 41)]
+        #[allow(non_upper_case_globals)]
+        /// EigenLayer-aligned genesis timestamp for rewards calculation.
+        pub static RewardsGenesisTimestamp: u32 = 0;
```
gonzamontiel, line 404: Same as prev comment

operator/runtime/testnet/src/configs/runtime_params.rs
gonzamontiel: Can't we get rid of this parameter entirely now that we should be using the `DatahavenServiceManagerAddress`. Same for the other runtimes.

operator/pallets/external-validators-rewards/src/lib.rs
```diff
@@ -308,7 +345,6 @@ pub mod pallet {
 
             // Mint tokens using the configurable handler
             let ethereum_sovereign_account = T::RewardsEthereumSovereignAccount::get();
-            let inflation_amount = T::EraInflationProvider::get();
             if let Err(err) =
                 T::HandleInflation::mint_inflation(&ethereum_sovereign_account, inflation_amount)
```
gonzamontiel: There's a possible corner/initial case in which the network is live, and some validators have non zero points, but the addresses are not configured properly yet (`WHAVETokenAddress` or `DatahavenServiceManagerAddress` are zero, for example).
In this case, `mint_inflation()` would still be called, but no message will make it to EigenLayer, so we would be indeed minting inflation with no-distribution. Maybe is unlikely because we need to configure those parameters before, but to be on the safe side we can also avoid minting if the addresses are mis-configured.

operator/runtime/common/src/rewards_adapter.rs
```diff
@@ -0,0 +1,925 @@
...
+        return None;
+    }
+
+    if remainder > 0 {
+        log::debug!(target: LOG_TARGET, "Reward distribution remainder (dust): {} tokens", remainder);
+    }
+
+    let calldata = encode_rewards_calldata(
+        whave_token_address,
+        &C::strategies_and_multipliers(),
```
gonzamontiel: Right now this is an empty vec, but this can easily change and break if the implementation doesn't respect the commment:

> /// EigenLayer requires `strategiesAndMultipliers` to be sorted by strategy address (ascending)
> /// with no duplicates. Multipliers must fit in `uint96`.

Not mandatory, but we could have a validation before sending the rewards that this is indeed sorted, etc.

gonzamontiel: Or we can also add a check in `DataHavenServiceManagerr.sol`, in `submitRewards()`, at the very beginning before summing the amounts.

operator/runtime/common/src/rewards_adapter.rs
```diff
@@ -0,0 +1,925 @@
...
+///
+/// # Arguments
+/// * `points` - List of (operator, points) tuples
+/// * `total_points` - Sum of all points
+/// * `inflation` - Total tokens to distribute
+///
+/// # Returns
+/// `Ok((operator_rewards, remainder))` where:
+/// * `operator_rewards` - Sorted list of (operator_address, amount) tuples
+/// * `remainder` - Dust amount from integer division truncation
```
gonzamontiel, line 233: Do we have a plan for this one? I know is very little, but over time maybe we want to do something since it's being minted? Just a thought for the future.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -232,6 +236,48 @@ contract DataHavenServiceManager is ServiceManagerBase, IDataHavenServiceManager
...
+            operatorSet, submissions
+        );
+
+        emit RewardsSubmitted(totalAmount, submission.operatorRewards.length);
+    }
+
+    /// @notice Sets the rewards initiator address (overrides deprecated base implementation)
+    /// @param newRewardsInitiator The new rewards initiator address
+    /// @dev Only callable by the owner
+    function setRewardsInitiator(
```
gonzamontiel, line 271: If for any reason this rewardsInitiator is misconfigured (or for example the zero address), every rewards submission tx from the bridge reverts with `OnlyRewardsInitiator()`. In this scenario `on_era_end` will still mint inflation and we will still mint gas from the failed `submitRewards()`. Maybe we should add an extra check here or in `submitRewards()`.

EDIT: this is not needed because we want to mint linearly (so on every on_era_end() )

contracts/test/RewardsSubmitter.t.sol
```diff
@@ -0,0 +1,262 @@
...
+        IRewardsCoordinatorTypes.OperatorDirectedRewardsSubmission memory submission =
+            _buildSubmission(rewardAmount, operator1);
+
+        // Warp to a time after the period ends
+        vm.warp(submission.startTimestamp + submission.duration + 1);
+
+        vm.prank(snowbridgeAgent);
+        vm.expectEmit(false, false, false, true);
+        emit IDataHavenServiceManagerEvents.RewardsSubmitted(rewardAmount, 1);
+        serviceManager.submitRewards(submission);
```
undercover-cactus, line 123: I don't see the operator validator being registered in this test. I am wrong to think that a validator should be register as a validator to get a reward ?

operator/runtime/common/src/rewards_adapter.rs
```diff
@@ -0,0 +1,925 @@
...
+pub enum RewardsAdapterError {
+    /// A strategy multiplier exceeds the maximum value for uint96.
+    InvalidMultiplier,
+    /// An arithmetic multiplication overflowed.
+    MultiplicationOverflow,
+    /// An arithmetic division by zero.
+    DivisionByZero,
+}
+
+sol! {
```
undercover-cactus, line 57: This is a really useful macro. It avoid writing painful encoding. Could you confirm it is working correctly in the runtime ? It doesn't rely on `std` ?

ahmadkaouk: Yes, it's working correctly and don't rely on `std`

operator/runtime/mainnet/src/configs/runtime_params.rs
ahmadkaouk: This will be used in a follow-up PR about the rewards distribution.

operator/runtime/testnet/src/configs/runtime_params.rs
ahmadkaouk: Fixed

operator/runtime/common/src/rewards_adapter.rs
ahmadkaouk: Fixed in [f0e55c3](https://github.com/datahaven-xyz/datahaven/pull/351/commits/f0e55c30c3d5ba4f65852fce92aae4bdee487ab8)

operator/pallets/external-validators-rewards/src/lib.rs
ahmadkaouk: Not really related to the changes in this PR, but as discussed, we should probably mint inflation consistently to ensure predictable inflation. On the other hand, to send the reward message, there is already a check of addresses. https://github.com/datahaven-xyz/datahaven/blob/d42a9283e2dd94e1c9719b7426be4ff93ad0c1ac/operator/runtime/common/src/rewards_adapter.rs#L150-L161

operator/runtime/common/src/rewards_adapter.rs
ahmadkaouk: Fixed in [e824906](https://github.com/datahaven-xyz/datahaven/pull/351/commits/e8249062a7b12b5137757632fbc7aa295fbc3f50). Remainder is transferred to the treasury account.

operator/runtime/common/src/rewards_adapter.rs
```diff
@@ -0,0 +1,929 @@
...
+use sp_std::vec::Vec;
+
+/// Default description for rewards submissions.
+pub const REWARDS_DESCRIPTION: &str = "DataHaven validator rewards";
+
+/// Log target for rewards adapter messages.
+const LOG_TARGET: &str = "rewards_adapter";
+
+/// Gas limit for the submitRewards call on Ethereum.
+pub const SUBMIT_REWARDS_GAS_LIMIT: u64 = 2_000_000;
```
undercover-cactus, line 44: How did you calculate the `SUBMIT_REWARDS_GAS_LIMIT` here ?

operator/runtime/common/src/rewards_adapter.rs
```diff
@@ -0,0 +1,929 @@
...
+        gas: SUBMIT_REWARDS_GAS_LIMIT,
+        value: 0,
+    }]
+    .try_into()
+    .ok()?;
+
+    Some(OutboundMessage {
+        origin: C::rewards_agent_origin(),
+        id: H256::from_low_u64_be(rewards_utils.era_index as u64).into(),
+        fee: 0,
```
undercover-cactus, line 214: Why the fee here is 0 ?

operator/runtime/common/src/rewards_adapter.rs
ahmadkaouk: I didn’t calculate the value. I simply passed a limit that is high enough to ensure the call has sufficient gas. If the gas limit is too low, the entire cross-chain message fails and cannot be retried.

operator/runtime/common/src/rewards_adapter.rs
ahmadkaouk: fee currently has no practical impact since the relayers are configured to skip fee check. https://github.com/datahaven-xyz/datahaven/blob/42ec577f15fbd0e62eda12d22e2df479510def19/test/configs/snowbridge/stagenet/solochain-relay.json#L31-L35
This will change once we enable fee check.

gonzamontiel (approved):
Nice work. I'm fine with merging it !

PR #356, opened 2025-12-15
fix: 🔨 Fix Kurtosis & Snowbridge relay configs for Fulu fork

stiiifff:


deploy/charts/relay/configs/beefy-relay.json
```diff
@@ -16,8 +16,9 @@
         }
     },
     "on-demand-sync": {
-        "max-tokens": 5,
-        "refill-amount": 1,
-        "refill-period": 3600
+        "asset-hub-channel-id": "0x0000000000000000000000000000000000000000000000000000000000000001",
```
gonzamontiel, line 19: Do we need this even though we use v2 and no channel is needed for us ?

PR #363, opened 2025-12-18
feat: 📈 Implement linear (non-compounding) inflation model

stiiifff:
## Summary

This PR replaces the percentage-based compounding inflation model with a **linear (non-compounding) inflation model** where a fixed amount of tokens is minted annually, regardless of current total supply.

### Key Changes

- **`ExternalRewardsEraInflationProvider`** now calculates per-era inflation from a fixed annual amount instead of a percentage of current total issuance
- New **`InflationAnnualAmount`** runtime parameter using the formula: `5_000_000 * HAVE * SUPPLY_FACTOR`
- Consistent configuration across all runtimes using `SUPPLY_FACTOR`

### Inflation Configuration

| Runtime | SUPPLY_FACTOR | Genesis Supply | Annual Inflation | Per-Era Inflation |
|---------|---------------|----------------|------------------|-------------------|
| **Mainnet** | 100 | 10B HAVE | 500M HAVE (5%) | ~342,231 HAVE |
| **Stagenet** | 1 | 100M HAVE | 5M HAVE (5%) | ~3,422 HAVE |
| **Testnet** | 1 | 100M HAVE | 5M HAVE (5%) | ~3,422 HAVE |

### Benefits

- **Predictable rewards**: Validators and stakers receive consistent emissions
- **Publicly auditable**: All emissions recorded on-chain
- **Non-compounding**: Same absolute amount minted each year (not percentage of growing supply)
- **Governance-upgradeable**: `InflationAnnualAmount` can be changed via runtime parameters

### Comparison: Before vs After

| Aspect | Before (Compounding) | After (Linear) |
|--------|---------------------|----------------|
| Formula | 5% × current_supply | Fixed 500M HAVE |
| Year 1 (10B supply) | 500M HAVE | 500M HAVE |
| Year 2 (10.5B supply) | 525M HAVE | 500M HAVE |
| Year 10 | ~814M HAVE | 500M HAVE |

## ⚠️ Breaking Changes ⚠️

- **Runtime parameter renamed**: `InflationTargetedAnnualRate` (Perbill) → `InflationAnnualAmount` (Balance)
  - Old: percentage-based rate applied to current total issuance
  - New: fixed annual amount in base units (wei)
- **`ExternalRewardsEraInflationProvider` type parameters changed**:
  - Removed: `Balances` (fungible::Inspect) and `AnnualRate` (Get<Perbill>)
  - Added: `AnnualAmount` (Get<u128>)
- **Inflation behavior change**: Inflation is now linear (fixed amount) instead of compounding (percentage of supply)


operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -350,10 +350,21 @@ pub mod dynamic_params {
...
+        /// - Publicly auditable emissions on the blockchain
+        /// - 5% of genesis supply, not 5% of current supply
+        ///
+        /// Formula: 5_000_000 * HAVE * SUPPLY_FACTOR
+        /// - Base: 5M HAVE annual inflation (5% of 100M base supply)
+        /// - Stagenet (SUPPLY_FACTOR=1): 5M HAVE annual (5% of 100M)
+        ///
+        /// The annual amount is divided equally across all eras in a year (~1461 eras with 6-hour eras).
+        /// Per-era inflation ≈ 3,422 HAVE (stagenet)
+        pub static InflationAnnualAmount: Balance = 5_000_000 * HAVE * SUPPLY_FACTOR;
```
gonzamontiel, line 367: We need to be careful with reusing indexes and changing types. This index `37` was a `Perbill` before, and I'm not sure, but if this is live on testnet, and was set as a parameter, we might be miss reading the value in bytes as a Balance now. Probably unimportant now, but a warning for the future

stiiifff: You're right. https://github.com/datahaven-xyz/datahaven/pull/304 is live on Testnet as part of [RT800](https://github.com/datahaven-xyz/datahaven/pull/304) but the dynamic param was not set (and there's no plan to do so) so there shouldn't be anything in storage. I think it should be fine to change this at this time, but it's a good point for future updates.

PR #367, opened 2025-12-19
fix: Use block authorship as liveness indicator for validator rewards

stiiifff:
## Summary

Use block authorship as direct proof of liveness for the 30% liveness component of validator rewards. Validators who author at least one block in a session are considered online and receive the full liveness bonus.

## Problem

The rewards pallet was checking validator liveness via ImOnline **after** the session had rotated - at which point ImOnline had already cleared its `AuthoredBlocks` storage. This caused all validators to appear offline, resulting in only ~70% of expected rewards being allocated (missing the 30% liveness bonus).

## Solution

Use **block authorship as the proxy for liveness**:

- A validator who authored at least one block is definitively online
- Liveness is determined directly in `award_session_performance_points` via `blocks_authored > 0`
- No dependency on external liveness checks (ImOnline)

### Rewards Formula

- **60%** Block authorship (proportional to blocks produced)
- **30%** Liveness (full bonus if authored ≥1 block, zero otherwise)
- **10%** Base reward (for being in the validator set)

### Files Changed

- `pallets/external-validators-rewards/src/lib.rs` - Core logic changes
- `pallets/external-validators-rewards/src/mock.rs` - Test mock updates
- `pallets/external-validators-rewards/src/tests.rs` - Test updates
- `runtime/{mainnet,testnet,stagenet}/src/configs/mod.rs` - Config updates

## Testing

- All 76 pallet tests pass
- Local testing should show correct points per session (e.g., 3200 points for 2 validators with 10 blocks)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,3 +73,6 @@ jobs:
     # Note: GITHUB_TOKEN is automatically available to reusable workflows
     with:
       image-tag: ${{ needs.docker-build-ci.outputs.image-tag }}
+    secrets:
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/task-e2e.yml
+++ b/.github/workflows/task-e2e.yml
@@ -21,6 +21,11 @@ on:
         description: "The tag of the docker image"
         required: true
         type: string
+    secrets:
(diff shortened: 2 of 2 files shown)


operator/pallets/external-validators-rewards/src/lib.rs
gonzamontiel, line 460: I think we need to update this documentation.

stiiifff: Good point, the doc comment was updated.

ahmadkaouk (approved):

PR #372, opened 2025-12-30
feat: :loud_sound: Add CPUs log at startup

ffarall:
Adds a log at the startup of the node to show the number of logical CPUs available to use for parallelism.


gonzamontiel (approved):
The comment is explicit, the change is harmless, if we find out that this is consistently the number of real cores, then we can remove.

PR #384, opened 2026-01-08
fix: update weights for RT910

gonzamontiel:
- [x] Updated weights for testnet
- [x] Updated weights for stagenet
- [x] Updated weights for mainnet


stiiifff (commented):
Overall LGTM. Just remove the extraneous `pallet_collective` weights, and review  the few recommendations in my comment.

operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -1107,34 +1107,36 @@ impl snowbridge_pallet_system_v2::Config for Runtime {
...
-            version: hex_literal::hex!("60000038"),
-            epoch: 0,
+            version: hex_literal::hex!("05000000"),
+            epoch: 80000000000,
         },
         fulu: Fork {
-            version: hex_literal::hex!("70000038"),
-            epoch: 0,
+            version: hex_literal::hex!("06000000"),
+            epoch: 90000000000,
```
ahmadkaouk: These values should remain unchanged as they are needed for the end-to-end tests. Otherwise, the Ethereum light client won’t work correctly. If you need to modify them for benchmarking, I suggest duplicating them and using those values under the `runtime-benchmarks` feature flag.

ahmadkaouk: fixed in 9ae92df13bb72fcd2a3a70bd4c3a89d28cc422e4

PR #389, opened 2026-01-12
refactor: Remove eigenlayer-middleware and flatten ServiceManagerBase

ahmadkaouk:
## Summary

  - Flatten `ServiceManagerBase` middleware layer directly into `DataHavenServiceManager`
  - Remove all unused EigenLayer integration code to keep the contract minimal
  - Fix access control on `deregisterOperatorFromOperatorSets` (was missing `onlyOwner`)

  ## Motivation

  The `ServiceManagerBase` from eigenlayer-middleware was designed for the old `AVSDirectory` model and included many generic functions DataHaven doesn't use. This refactor:

  - Reduces code complexity and contract size
  - Removes ~200 lines of unused code
  - Makes the codebase easier to audit and maintain
  - Keeps only what DataHaven actually needs

  ## Changes

  ### Architecture
  Before: DataHavenServiceManager → ServiceManagerBase → ServiceManagerBaseStorage → OwnableUpgradeable
  After:  DataHavenServiceManager → OwnableUpgradeable, IAVSRegistrar, IDataHavenServiceManager

  ### Removed (unused)
  - `IServiceManager` and `IServiceManagerUI` interfaces (old AVSDirectory model)
  - `ServiceManagerBase` and `ServiceManagerBaseStorage` middleware
  - `PermissionController` integration (5 proxy functions)
  - `createOperatorSets()` - only needed at initialization
  - `avs()` - never called

  ### Kept (with fixes)
  - `deregisterOperatorFromOperatorSets()` - added `onlyOwner` modifier (security fix)
  - `updateAVSMetadataURI()` - needed for EigenLayer registration

  ### Files Deleted
  - `src/interfaces/IServiceManager.sol`
  - `src/interfaces/IServiceManagerUI.sol`
  - `src/middleware/ServiceManagerBase.sol`
  - `src/middleware/ServiceManagerBaseStorage.sol`
  - `test/mocks/ServiceManagerMock.sol`
  - `test/ServiceManagerBase.t.sol`

  ## Test Plan

  - [x] `forge build` passes
  - [x] `forge test` - all 10 tests pass
  - [x] Contract bindings regenerated
  - [x] State diff regenerated


stiiifff (changes requested):
Things to add:

1. The `registerOperator` & `deregisterOperator` functions should enforce that calls are made by EL AllocationManager only (otherwise anyone could call `registerOperator` and set an allowed operator's solochain address with a malicious intent).
2. There's no validation for zero addresses. This should be added across the board to prevent issues.
3. We should add a storage layout test that verifies slots don't change between versions. This will prevent accidental storage layout changes in the future.

Points 2 & 3 can be added in a separate PR.

ahmadkaouk:
> Things to add:
> 
> 1. The `registerOperator` & `deregisterOperator` functions should enforce that calls are made by EL AllocationManager only (otherwise anyone could call `registerOperator` and set an allowed operator's solochain address with a malicious intent).
> 2. There's no validation for zero addresses. This should be added across the board to prevent issues.
> 3. We should add a storage layout test that verifies slots don't change between versions. This will prevent accidental storage layout changes in the future.
> 
> Points 2 & 3 can be added in a separate PR.

Point 1 fixed in [ff21aab](https://github.com/datahaven-xyz/datahaven/pull/389/commits/ff21aab8f83be204363c4b6e3d772575af3bc468)
For 2 and 3 I am already working on a followup PR for some additional fixes/improvements to the contract.

PR #394, opened 2026-01-14
test: storagehub e2e

undercover-cactus:
## Summary

Add Storage Hub basic end to end test. This PR also include some fixes to allow Storage Hub node and datahaven node to run on the same network (local chain). Before that one was running on dev and the other one on the local chain.

## What changed

* Added `storagehub.test.ts` e2e  test. In this file we explicitly start the storagehub node using the launch function already used in the CI
* Added Storage Hub backend the flow so it can be used in the e2e test
* Fix the `--chain local` vs `--chain dev` issue. The storagehub nodes were started on the dev network and therefore they were never syncing with the datahaven node
* Fix  the folder permission issue in the CI by fixing the folder name
* Added StorageHub javascript lib

### DIFF
--- a/test/e2e/suites/storagehub.test.ts
+++ b/test/e2e/suites/storagehub.test.ts
@@ -4,7 +4,7 @@
  * Tests the uploading a file to storage through Datahaven
  *
  * Prerequisites:
- * - DataHaven network with StoraheHub service running
+ * - DataHaven network with StorageHub service running
  * - Storage hub MSP and BSP
  */
 import "@storagehub/api-augment";
@@ -126,7 +126,6 @@ describe("test uploading file to storage hub", () => {
     const aliceSigner = getEvmEcdsaSigner(SUBSTRATE_FUNDED_ACCOUNTS.ALITH.privateKey);
     const mspResult = await call.signAndSubmit(aliceSigner);
     expect(mspResult.ok).toBeTrue();
-    console.log(mspResult);
   }, 30000);
 
   it("Send a request", async () => {
@@ -212,29 +211,18 @@ describe("test uploading file to storage hub", () => {
     console.log("issueStorageRequest() txReceipt:", receipt);
 
     // Authenticate with the backend via SIWE and upload the file
-    const mspClient = await MspClient.connect({ baseUrl: backendUrl }, async () =>
-      sessionToken
-        ? ({ token: sessionToken, user: { address: account.address } } as const)
-        : undefined
-    );
+    let sessionRef: { token: string; user: { address: string } } | undefined;
+    const sessionProvider = async () => sessionRef;
+    const mspClient = await MspClient.connect({ baseUrl: backendUrl }, sessionProvider);
 
     const domain = new URL(backendUrl).host;
     const siweSession = await mspClient.auth.SIWE(walletClient, domain, backendUrl);
     const sessionToken = (siweSession as { token: string }).token;
     expect(sessionToken).toBeDefined();
-
-    // const uploadReceipt = await mspClient.files.uploadFile(
-    //   bucketId,
-    //   fileKey.toHex(),
-    //   await fileManager.getFileBlob(),
-    //   account.address,
-    //   location
-    // );
-    // expect(uploadReceipt.status).toBe("upload_successful");
   }, 60000);
 
   afterAll(async () => {
     // Delete all the containers started by this test suite
-    await $`docker container stop $(docker container ls -q --filter name=${networkId})`;
+    await $`docker container rm -f $(docker container ls -q --filter name=${networkId})`;
   });
 });


test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+    const receipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+    if (receipt.status !== "success") {
+      throw new Error(`Storage request failed: ${txHash}`);
+    }
+    console.log("issueStorageRequest() txReceipt:", receipt);
+
+    // Authenticate with the backend via SIWE and upload the file
+    const mspClient = await MspClient.connect({ baseUrl: backendUrl }, async () =>
+      sessionToken
```
gonzamontiel: Possible race condition since sessionToken si defined after this callback is defined, please move the declaration above this line.

test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        ? ({ token: sessionToken, user: { address: account.address } } as const)
+        : undefined
+    );
+
+    const domain = new URL(backendUrl).host;
+    const siweSession = await mspClient.auth.SIWE(walletClient, domain, backendUrl);
+    const sessionToken = (siweSession as { token: string }).token;
+    expect(sessionToken).toBeDefined();
+
+    // const uploadReceipt = await mspClient.files.uploadFile(
```
gonzamontiel: I'm not a fan of leaving commented code, can we remove / explain ?

test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    //   await fileManager.getFileBlob(),
+    //   account.address,
+    //   location
+    // );
+    // expect(uploadReceipt.status).toBe("upload_successful");
+  }, 60000);
+
+  afterAll(async () => {
+    // Delete all the containers started by this test suite
+    await $`docker container stop $(docker container ls -q --filter name=${networkId})`;
```
gonzamontiel: We are not deleting the containers here, so the next run can hit “name already in use” error

gonzamontiel: You can mimic the pattern used in `network/index.ts` inside `createCleanupFunction()`

test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
+/**
+ * StorageHub E2E Tests
+ *
+ * Tests the uploading a file to storage through Datahaven
+ *
+ * Prerequisites:
+ * - DataHaven network with StoraheHub service running
```
gonzamontiel:
```suggestion
 * - DataHaven network with StorageHub service running
```

test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    const call = await dhApi.tx.FileSystem.create_bucket({
+      msp_id,
+      name: Binary.fromText("bucket"),
+      private: false,
+      value_prop_id: value_prop_id[0].id
+    });
+    const aliceSigner = getEvmEcdsaSigner(SUBSTRATE_FUNDED_ACCOUNTS.ALITH.privateKey);
+    const mspResult = await call.signAndSubmit(aliceSigner);
+    expect(mspResult.ok).toBeTrue();
+    console.log(mspResult);
```
gonzamontiel: Is this is a formal log, please use `logger` instead, otherwise let's remove.

test/e2e/suites/storagehub.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        }) as ReadableStream<Uint8Array>
+    });
+
+    // Compute fingerprint and file key from the file metadata
+    const registry = new TypeRegistry();
+    const account = privateKeyToAccount(SUBSTRATE_FUNDED_ACCOUNTS.ALITH.privateKey);
+    const owner = registry.createType("AccountId20", account.address);
+    const bucketIdH256 = registry.createType("H256", bucketId);
+    const fingerprint = await fileManager.getFingerprint();
+    const _fileKey = await fileManager.computeFileKey(owner, bucketIdH256, location);
```
gonzamontiel, line 171: This seems unused

test/e2e/suites/storagehub.test.ts
undercover-cactus: I can't move the line around because we need the `mspClient` defined to get the session. The function is callback.

I have changed things a bit so it doesn't look like it will lead to a race condition. However we are stuck with how the SDK works.

test/e2e/suites/storagehub.test.ts
undercover-cactus: Yeah I removed it. It was to go further with the test. There is already a lot with this PR so I left this on the side.

test/e2e/suites/storagehub.test.ts
undercover-cactus: I have changed the command to remove the container. I was just stopping them for debugging. (Once the container is removed you can't access the logs)

test/e2e/suites/storagehub.test.ts
undercover-cactus: Fixed

test/e2e/suites/storagehub.test.ts
undercover-cactus: I have removed it.

test/e2e/suites/storagehub.test.ts
undercover-cactus: It is unused. However I would not remove it. There is still room to add more tests and I want to keep it as the `fileKey` value is needed in some other process. Also it shows the difference with the `fingerprint` value. They are not the same and can be confusing.

I believe a unused variable in a test file can be tolerated.

PR #395, opened 2026-01-15
refactor(contracts): Harden DataHavenServiceManager with input validation and code cleanup

ahmadkaouk:
## Summary

- Add zero address validation across all functions that accept address parameters to prevent misconfiguration
- Fix race condition in `buildNewValidatorSetMessage()` that could cause reverts during validator deregistration
- Refactor contract for improved readability and reduced code duplication
- Update AVS metadata URL to point to the correct hosted JSON file

### Security & Validation
- Add `ZeroAddress` error and validate all address inputs in `initialize`, `setRewardsInitiator`, `setSnowbridgeGateway`, `addValidatorToAllowlist`, `registerOperator`, and `updateSolochainAddressForValidator`
- Fix race condition: filter out zero solochain addresses in `buildNewValidatorSetMessage()` to prevent reverts when a validator is mid-deregistration

### Refactoring
- Replace verbose `if/revert` patterns with `require` statements for consistency
- Inline single-use internal functions (`_createDataHavenOperatorSets`, `_setRewardsInitiator`)
- Consolidate duplicate error types into single `ZeroAddress` error
- Rename `initialise` → `initialize` to maintain consistency with the transparent upgradability pattern
- Optimize validator set message encoding by removing redundant wrapper function

### Observability
- Add `SolochainAddressUpdated` event for tracking validator address changes

### Cleanup
- Remove unused remappings from `foundry.toml`
- Fix typo in metadata description


contracts/src/DataHavenServiceManager.sol
```diff
@@ -105,31 +104,34 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
-
     /// @inheritdoc IDataHavenServiceManager
-    function initialise(
+    function initialize(
         address initialOwner,
         address _rewardsInitiator,
         IStrategy[] memory validatorsStrategies,
         address _snowbridgeGatewayAddress
     ) public virtual initializer {
+        require(initialOwner != address(0), ZeroAddress());
```
stiiifff, line 114: Shouldn't the `_rewardsInitiator` & `_snowbridgeGatewayAddress` also be checked ?

contracts/src/DataHavenServiceManager.sol
```diff
@@ -187,24 +200,14 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
-
-        if (!validatorsAllowlist[operator]) {
-            revert OperatorNotInAllowlist();
-        }
-
+        require(avsAddress == address(this), IncorrectAVSAddress());
+        require(operatorSetIds.length == 1, CantRegisterToMultipleOperatorSets());
+        require(operatorSetIds[0] == VALIDATORS_SET_ID, InvalidOperatorSetId());
+        require(validatorsAllowlist[operator], OperatorNotInAllowlist());
         require(data.length == 20, "Invalid solochain address length");
```
stiiifff: This check should probably use a custom error instead of string literal, to be consistent with other error checks (that use custom errors).

contracts/src/DataHavenServiceManager.sol
```diff
@@ -105,31 +104,34 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
         address _rewardsInitiator,
         IStrategy[] memory validatorsStrategies,
         address _snowbridgeGatewayAddress
     ) public virtual initializer {
+        require(initialOwner != address(0), ZeroAddress());
+
         __Ownable_init();
         _transferOwnership(initialOwner);
-        _setRewardsInitiator(_rewardsInitiator);
+        rewardsInitiator = _rewardsInitiator;
```
stiiifff, line 120: Should `RewardsInitiatorSet` be emitted during initialization ?

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Fixed in [66e56f0](https://github.com/datahaven-xyz/datahaven/pull/395/commits/66e56f0fb3dac3ea1009fe0e89495d84744b5fd6)

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Fixed in [87eb40f](https://github.com/datahaven-xyz/datahaven/pull/395/commits/87eb40fe53c83799ff4f3d624f12ae6eb3b48390)

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Fixed in [26975b4](https://github.com/datahaven-xyz/datahaven/pull/395/commits/26975b42cf9914c5c6211fc23fb5dab25c7dfa60)

contracts/src/DataHavenServiceManager.sol
```diff
@@ -187,24 +203,14 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
-        require(data.length == 20, "Invalid solochain address length");
-        validatorEthAddressToSolochainAddress[operator] = address(bytes20(data));
+        require(avsAddress == address(this), IncorrectAVSAddress());
+        require(operatorSetIds.length == 1, CantRegisterToMultipleOperatorSets());
+        require(operatorSetIds[0] == VALIDATORS_SET_ID, InvalidOperatorSetId());
+        require(validatorsAllowlist[operator], OperatorNotInAllowlist());
+        require(data.length == 20, InvalidSolochainAddressLength());
+        address solochainAddress = address(bytes20(data));
+        require(solochainAddress != address(0), ZeroAddress());
+        validatorEthAddressToSolochainAddress[operator] = solochainAddress;
```
undercover-cactus, line 213: Only a recommendation here. Because this a callback function to add extra verification when registering operator, I think we should change the `validatorEthAddressToSolochainAddress` mapping somewhere else.

We could create a `registerOperatorFromAVS` function so we register using our AVS contract and not directly with the allocation manager. We can also introduce the update `validatorEthAddressToSolochainAddress[operator] = solochainAddress;` in this new function. The cons would be extra gas fee cost. What do you think ?

contracts/src/DataHavenServiceManager.sol
```diff
@@ -215,17 +221,9 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
-        if (operatorSetIds.length != 1) {
-            revert CantDeregisterFromMultipleOperatorSets();
-        }
-
-        if (operatorSetIds[0] != VALIDATORS_SET_ID) {
-            revert InvalidOperatorSetId();
-        }
+        require(avsAddress == address(this), IncorrectAVSAddress());
+        require(operatorSetIds.length == 1, CantDeregisterFromMultipleOperatorSets());
+        require(operatorSetIds[0] == VALIDATORS_SET_ID, InvalidOperatorSetId());
```
undercover-cactus, line 226: Looks better like this :+1:

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Requiring users to register through the AllocationManager and then separately call the AVS to set the solochain address has two concerns:
  1. UX degradation: Operators would need to coordinate two transactions instead of one, adding friction to the registration flow.
  2. Inconsistent state: This creates a window where an operator is registered but has no solochain address. We would need to handle this partial state throughout the codebase.

Given these trade-offs, I am leaning to the current atomic approach.

PR #398, opened 2026-01-16
fix: resolve forge build warnings

ahmadkaouk:
### Configuration
- Remove deprecated `deny_warnings` config key from foundry.toml
- Add global `[lint]` config to suppress naming convention warnings for AVS/EL/ERC patterns (`mixed-case-function`, `mixed-case-variable`)

### DataHavenServiceManager Refactoring
- Rename immutable variables to SCREAMING_SNAKE_CASE (`_allocationManager` → `_ALLOCATION_MANAGER`, `_rewardsCoordinator` → `_REWARDS_COORDINATOR`)
- Wrap modifier logic in internal functions (`_checkRewardsInitiator`, `_checkValidator`, `_checkAllocationManager`) to reduce contract size
- Add `_toAddress` helper with assembly for safe bytes-to-address conversion

### Safe Typecasting
- Replace direct typecasts with OpenZeppelin's SafeCast library in deploy scripts and test utilities
- Use `.toUint32()`, `.toUint64()`, `.toUint160()` for overflow-protected conversions
- Replace `bytes32("wrong origin")` string cast with hex literal in test deployer

### Code Cleanup
- Remove 25+ unused imports across script and test files
- Convert plain imports to named imports for better clarity
- Use `SafeERC20.safeTransfer()` for token transfers in tests
- Change `view` to `pure` where appropriate

## Test plan

- [x] `forge build` completes with no warnings
- [x] `forge test` passes all 10 tests

### DIFF
--- a/contracts/src/DataHavenServiceManager.sol
+++ b/contracts/src/DataHavenServiceManager.sol
@@ -221,7 +221,6 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
         require(operatorSetIds.length == 1, CantRegisterToMultipleOperatorSets());
         require(operatorSetIds[0] == VALIDATORS_SET_ID, InvalidOperatorSetId());
         require(validatorsAllowlist[operator], OperatorNotInAllowlist());
-        require(data.length == 20, InvalidSolochainAddressLength());
         validatorEthAddressToSolochainAddress[operator] = _toAddress(data);
 
         emit OperatorRegistered(operator, operatorSetIds[0]);
@@ -374,31 +373,6 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
     }
 
     // ============ Internal Functions ============
-
-    /**
-     * @notice Creates the initial operator set for DataHaven in the AllocationManager.
-     * @dev This function should be called during initialisation to set up the required operator set.
-     */
-    function _createDataHavenOperatorSets(
-        IStrategy[] memory validatorsStrategies
-    ) internal {
-        IAllocationManagerTypes.CreateSetParams[] memory operatorSets =
-            new IAllocationManagerTypes.CreateSetParams[](1);
-        operatorSets[0] = IAllocationManagerTypes.CreateSetParams({
-            operatorSetId: VALIDATORS_SET_ID, strategies: validatorsStrategies
-        });
-        _ALLOCATION_MANAGER.createOperatorSets(address(this), operatorSets);
-    }
-
-    /**
-     * @notice Internal function to set the rewards initiator
-     * @param _rewardsInitiator The new rewards initiator address
-     */
-    function _setRewardsInitiator(
-        address _rewardsInitiator
-    ) internal {
-        rewardsInitiator = _rewardsInitiator;
-    }
 
     /**
      * @notice Safely converts a 20-byte array to an address


contracts/src/DataHavenServiceManager.sol
```diff
@@ -121,15 +135,15 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
+        _ALLOCATION_MANAGER.updateAVSMetadataURI(address(this), DATAHAVEN_AVS_METADATA);
 
         // Create the operator set for the DataHaven service.
         IAllocationManagerTypes.CreateSetParams[] memory operatorSets =
             new IAllocationManagerTypes.CreateSetParams[](1);
         operatorSets[0] = IAllocationManagerTypes.CreateSetParams({
             operatorSetId: VALIDATORS_SET_ID, strategies: validatorsStrategies
         });
-        _allocationManager.createOperatorSets(address(this), operatorSets);
+        _ALLOCATION_MANAGER.createOperatorSets(address(this), operatorSets);
```
stiiifff, line 146: Looks like the same logic as in the `__createDataHavenOperatorSets` internal function .. that is never called. Probably replace this block with a call to the internal func.

contracts/src/DataHavenServiceManager.sol
stiiifff, line 134: Looks like `_setRewardsInitiator` is never called .. either we should remove it, or use it everywhere when setting the `rewardsInitiator`.

stiiifff, line 254: We should probably consider having a function that allows to add several validators at once, to be more gas-efficient.

contracts/src/DataHavenServiceManager.sol
```diff
@@ -208,9 +222,7 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
         require(operatorSetIds[0] == VALIDATORS_SET_ID, InvalidOperatorSetId());
         require(validatorsAllowlist[operator], OperatorNotInAllowlist());
         require(data.length == 20, InvalidSolochainAddressLength());
```
gonzamontiel: This is nice but `_toAddress()` already includes the same check, to be consistent I would only check it in one place (proably the most atomic place is `_toAddress()`). We can rename to `_to20BytesAddress()`, though.

contracts/script/deploy/DeployParams.s.sol
```diff
@@ -5,8 +5,11 @@ import {EmptyContract} from "eigenlayer-contracts/src/test/mocks/EmptyContract.s
 import {Config} from "./Config.sol";
 import {Script} from "forge-std/Script.sol";
 import {TestUtils} from "../../test/utils/TestUtils.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 contract DeployParams is Script, Config {
+    using SafeCast for uint256;
```
gonzamontiel, line 11: This is great, but now the use of longer/wrong addresses will revert instead of just truncating. We should pay attention to that.

gonzamontiel (approved):
Looks good, I only left a couple of comments.

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Yes, I agree. I will remove the redundant length check and keep it in the `_toAddress` function.

ahmadkaouk: Fixed in 5735642cada3ef4187116d4a8ad59c9639e512e3

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Yes, but I’ll leave this to a another PR since it’s beyond the scope of this one.

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Fixed in 5735642cada3ef4187116d4a8ad59c9639e512e3. I removed `_createDataHavenOperatorSets`.

contracts/src/DataHavenServiceManager.sol
ahmadkaouk: Fixed in 5735642cada3ef4187116d4a8ad59c9639e512e3

contracts/foundry.toml
```diff
@@ -115,4 +113,11 @@
     anvil = "http://localhost:8545"
 
 # [etherscan]
-#     mainnet = { key = "${ETHERSCAN_API_KEY}" }
\ No newline at end of file
+#     mainnet = { key = "${ETHERSCAN_API_KEY}" }
```
undercover-cactus, line 116: Should we remove this commented line ?

contracts/src/DataHavenServiceManager.sol
```diff
@@ -72,35 +72,49 @@ contract DataHavenServiceManager is OwnableUpgradeable, IAVSRegistrar, IDataHave
...
     modifier onlyAllocationManager() {
-        require(msg.sender == address(_allocationManager), OnlyAllocationManager());
+        _checkAllocationManager();
         _;
     }
 
+    function _checkRewardsInitiator() internal view {
+        require(msg.sender == rewardsInitiator, OnlyRewardsInitiator());
+    }
+
```
undercover-cactus, line 94: Are those function being used in other places ? If they are being called only in the `onlyRewardsInitiator` modifier it might not be worth. It impact readability.

Same with `_checkValidator` and `_checkAllocationManager`

contracts/test/RewardsSubmitter.t.sol
```diff
@@ -46,7 +38,7 @@ contract RewardsSubmitterTest is AVSDeployer {
         serviceManager.setRewardsInitiator(snowbridgeAgent);
 
         // Fund the service manager with reward tokens
-        rewardToken.transfer(address(serviceManager), 100000e18);
+        IERC20(address(rewardToken)).safeTransfer(address(serviceManager), 100000e18);
```
undercover-cactus, line 41: What is the difference between `transfer` and `safeTransfer` ?

PR #402, opened 2026-01-20
fix: 🔧 do not pop slashes on failure

gonzamontiel:
## Summary
This PR introduces a fix in cases where processing a slash fails for different reasons, after being popped out from the `UnreportedSlashesQueue`.

## Problem
`UnreportedSlashesQueue::<T>::mutate(|queue| { ... pop_front() ... })` pops a slash to process and writes the mutated queue back to storage after the closure finishes. If any of these later steps fail:
- `T::SendMessage::build(...) `returns None
- `T::SendMessage::validate(...)` returns Err
- `T::SendMessage::deliver(...)` returns Err

the function exits early without re-adding the popped items, and since this runs from `on_initialize`, there’s no outer error propagation that would “roll back” the earlier storage write. So those slashes are dropped and never retried.

## Fix
Use `UnreportedSlashesQueue::<T>::try_mutate(...)` and perform the dequeue + processing in an atomic way. If delivery fails, an error is returned and the queue mutation is not committed.

## Caveats
The assumption is that something temporary prevented the slash to be processed, so in next sessions it will succeed. If this doesn't happen then the slash will live in storage forever, but at least it won't be gone.


undercover-cactus (approved):

operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -584,61 +584,71 @@ impl<T: Config> Pallet<T> {
...
-        Some(slashes_count)
+                let message_id = T::SendMessage::deliver(ticket)
+                    .map_err(|e| {
+                        log::error!(
+                            target: "ext_validators_slashes",
+                            "Failed to deliver outbound message: {:?}",
+                            e
+                        );
+                    })
+                    .map_err(|_| ())?;
```
ahmadkaouk:
```suggestion
                let message_id = T::SendMessage::deliver(ticket).map_err(|e| {
                    log::error!(
                        target: "ext_validators_slashes",
                        "Failed to deliver outbound message: {:?}",
                        e
                    );
                })?;
```
Same here.

operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -584,61 +584,71 @@ impl<T: Config> Pallet<T> {
...
+                // Validate and deliver the message
+                let ticket = T::SendMessage::validate(outbound)
+                    .map_err(|e| {
+                        log::error!(
+                            target: "ext_validators_slashes",
+                            "Failed to validate outbound message: {:?}",
+                            e
+                        );
+                    })
+                    .map_err(|_| ())?;
```
ahmadkaouk:
```suggestion
                let ticket = T::SendMessage::validate(outbound).map_err(|e| {
                    log::error!(
                        target: "ext_validators_slashes",
                        "Failed to validate outbound message: {:?}",
                        e
                    );
                })?;

```
Same as above.

operator/pallets/external-validator-slashes/src/lib.rs
```diff
@@ -584,61 +584,71 @@ impl<T: Config> Pallet<T> {
...
+                let slashes_count = slashes_to_send.len() as u32;
+
+                let outbound = T::SendMessage::build(&slashes_to_send, era_index)
+                    .ok_or_else(|| {
+                        log::error!(
+                            target: "ext_validators_slashes",
+                            "Failed to build outbound message"
+                        );
+                    })
+                    .map_err(|_| ())?;
```
ahmadkaouk:
```suggestion
                let outbound = T::SendMessage::build(&slashes_to_send, era_index)
                    .ok_or_else(|| {
                        log::error!(
                            target: "ext_validators_slashes",
                            "Failed to build outbound message"
                        );
                    })?;
```
`.map_err(|_| ())` call is redundant and can be removed.

gonzamontiel:
> Nice catch! However, I have some concerns about the current implementation:
> 
> 1. Deterministic failures (e.g., payload size constraints, invalid slash data...) in `process_slashes_queue` can permanently stall the queue. Any failure from build, validate, or deliver returns Err and rolls back the storage mutation. That means the same head-of-queue slashes are retried every block, preventing later slashes from being processed. We could create a queue for failed slashes that we attempt to retry.

This has been addressed by rotating failed items to the back of the queue on failure (e.g., a,b,c → b,c,a), ensuring the queue continues making forward progress even if a specific slash fails repeatedly. Additionally, message construction/validation/delivery is wrapped in `with_transaction`, so any storage side effects during send are rolled back on failure while still preserving queue liveness.

> 2. In case of a failure we are mapping the error to `None`. The `on_initialize` treats None as zero slashes processed, so persistent failures under-account weight since failed processing still consume runtime resources.

This shouldn't be the case anymore. `on_initialize` will charge `T::WeightInfo::process_slashes_queue(p)` with 
p = attempted count (even on persistent failures).

PR #416, opened 2026-01-27
fix: set username grace period and deposit

gonzamontiel:
### Summary
Set username deletion to use a 30‑day grace period (in blocks) and added a non‑zero username deposit, both based on the Moonbeam's runtimes. This makes username unbinding wait before deletion and makes authority‑issued usernames non‑free, mitigating the sybil vector while aligning with existing identity config patterns.

### Changes
- Configures for `pallet_identity`
  - `UsernameGracePeriod = 30 * DAYS`
  - `UsernameGracePeriod = deposit(0, MaxUsernameLength::get())`


operator/runtime/mainnet/src/configs/mod.rs
```diff
@@ -626,8 +628,8 @@ impl pallet_identity::Config for Runtime {
     type MaxSuffixLength = MaxSuffixLength;
     type MaxUsernameLength = MaxUsernameLength;
     type WeightInfo = pallet_identity::weights::SubstrateWeight<Runtime>;
-    type UsernameDeposit = ();
-    type UsernameGracePeriod = ();
+    type UsernameDeposit = UsernameDeposit;
```
ahmadkaouk, line 631: The documentation explicitly states that we need to perform a proper migration when changing the value of `UsernameDeposit`.
https://github.com/paritytech/polkadot-sdk/blob/5fd7bb53b86f381ea42c67cf3da04622aa3a9953/substrate/frame/identity/src/lib.rs#L190-L193

I’ve checked how the pallet handles deposits, and it actually stores the actual deposit amount associated with each username. Therefore, technically, increasing the deposit without migrating won’t break refunds. However, I still wonder if a migration is necessary here just to be safe.

gonzamontiel: I mean the concern is valid. But changing `UsernameDeposit` only affects future calls to `set_username_for` that use use_allocation = false. Existing usernames keep their originally reserved amount and are refunded using that stored amount. So a migration will be needed only if we want to retroactively change a user's deposit. I don't think this is a common thing, and I checked on Moonbeam and they don't include a migration for that.

PR #418, opened 2026-01-28
test: launch backend in e2e tests and cli

undercover-cactus:
## Summary

We are now launching the MSP backend when starting stpragehub services. In this PR, we also fix the MSP and BSP node configuration and register it with the correct keys.

## What changed

* Added a launch Backend MSP function that is called when launching storage hub services
* Fix the wrong genesis error message in storagehub node by removing the `--chain dev` flags (so it can be launch of the same network as our local datahaven nodes).
* Use the correct keys to register MSP and BSP. We were injecting different keys that the one we used for MSP and BSP registration leading to the MSP and BSP node to never fully register as storage providers.


test/launcher/storagehub-docker.ts
```diff
@@ -109,48 +113,27 @@ export const getPostgresUrl = (networkId: string): string => {
...
   containerName: string,
-  seed: string,
-  derivation: string
+  secretKey: string
 ): Promise<void> => {
-  logger.info(`🔑 Injecting key ${derivation} into ${containerName}...`);
+  logger.info(`🔑 Injecting key into ${containerName}...`);
 
-  const suri = `${seed}${derivation}`;
+  // const suri = `${seed}${derivation}`;
```
gonzamontiel: If we stopped supporting seed + derivation path, then let's remove this line.

test/launcher/storagehub-docker.ts
```diff
@@ -109,48 +113,27 @@ export const getPostgresUrl = (networkId: string): string => {
...
- * For local development with Docker, nodes on the same network can discover each other
- * via mDNS (--discover-local flag), so explicit bootnodes are optional.
- *
- * To use explicit bootnodes, we'd need to extract the peer ID from the validator node,
- * which requires querying the RPC endpoint. For simplicity in local dev, we skip this.
- *
- * @param containerName - Name of the validator container (e.g., datahaven-alice-cli-launch)
- * @returns Multiaddress string for bootnode, or empty string to skip bootnodes
- */
-export const getBootnodeAddress = async (containerName: string): Promise<string> => {
```
gonzamontiel, line 147: It's true that it makes no sense to have empty bootnodes, this was meant for extending it to support `bun cli deploy` in kubernetes, where local discovery won't work. But as long as we only need this for local testing, it's fine. It can be extended later.

test/launcher/storagehub-docker.ts
```diff
@@ -507,6 +446,60 @@ export const launchFishermanNode = async (
...
+  const command: string[] = [
+    "docker",
+    "run",
+    "-d",
+    "--name",
+    containerName,
+    "--network",
+    dockerNetworkName,
+    "-p",
+    `${apiPort}:8080`,
```
gonzamontiel, line 483: We are harcoding the external port here (`apiPort`), this may collide with other services, we can use dynamic ports intead.

test/scripts/fund-providers.ts
gonzamontiel: Not your change, but I beliece the commend should say Dorothy ? Since bsp account is Dorothy and not Eve.

test/launcher/storagehub-docker.ts
```diff
@@ -375,8 +331,6 @@ export const launchIndexerNode = async (
     "-p",
     `${wsPort}:${DEFAULT_SUBSTRATE_WS_PORT}`,
     options.datahavenImageTag,
-    "--chain",
-    "dev",
```
gonzamontiel, line 379: In fact to avoid future problems, it would be even better to specify `--chain local`, given that we are reducing the scope of this nodes to a local launch.

test/launcher/storagehub-docker.ts
undercover-cactus: I don't think it will be an issue. We are also hardcoding the external ports for all the other services being launched. We just need to make sure to reserve this port for the Backend MSP.

test/scripts/fund-providers.ts
undercover-cactus: Yeah those logs are not right. And we are not using the derivation path because the keys that we are using were not generated that way... So it is confusing.

test/launcher/storagehub-docker.ts
undercover-cactus: Yep! I always agree with explicit configuration. It always makes thing clearer.

test/launcher/storagehub-docker.ts
undercover-cactus: done

test/launcher/storagehub-docker.ts
```diff
@@ -109,48 +113,25 @@ export const getPostgresUrl = (networkId: string): string => {
...
   try {
-    await $`docker exec ${containerName} datahaven-node key insert --base-path /data --chain dev --key-type bcsv --scheme ecdsa --suri ${suri}`.nothrow();
-    logger.success(`Key ${derivation} injected successfully`);
+    await $`docker exec ${containerName} datahaven-node key insert --base-path /data --key-type bcsv --scheme ecdsa --suri ${secretKey}`.nothrow();
+    logger.success("Key injected successfully");
   } catch (error) {
-    logger.error(`Failed to inject key ${derivation}: ${error}`);
+    logger.error(`Failed to inject key : ${error}`);
     throw error;
   }
```
ahmadkaouk, line 132: We are using the `.nothrow()` method to not throw an exception when the command fails. However, the code wraps it in a try-catch expecting to catch errors. The catch block will never execute for command failures. I would remove the `no throw` method and the `try-catch` block and let the exception propagate.

test/launcher/storagehub-docker.ts
```diff
@@ -303,19 +270,13 @@ export const launchBspNode = async (
...
-
   logger.debug(`Executing: ${command.join(" ")}`);
   await $`sh -c "${command.join(" ")}"`.nothrow();
 
   await waitForContainerToStart(containerName);
 
   // Inject key (using Dave instead of Eve for BSP)
-  const seed = "bottom drive obey lake curtain smoke basket hold race lonely fit walk";
-  await injectStorageHubKey(containerName, seed, "//Dave");
+  await injectStorageHubKey(containerName, SUBSTRATE_FUNDED_ACCOUNTS.DOROTHY.privateKey);
```
ahmadkaouk: the node is named `bsp-dave` (passed in the flag) and the comment says “using Dave instead of Eve,” but the key injected is `SUBSTRATE_FUNDED_ACCOUNTS.DOROTHY`.

test/launcher/storagehub-docker.ts
```diff
@@ -507,6 +452,63 @@ export const launchFishermanNode = async (
...
+    "RUST_LOG=info",
+    backendMSPImage,
+    "--chain",
+    "local",
+    "--log-format",
+    "text",
+    "--database-url",
+    postgresUrl,
+    "--rpc-url",
+    `ws://${containerNameMSP}:9944`,
```
ahmadkaouk: Nit! For the sake of consistency:
```suggestion
    `ws://${containerNameMSP}:DEFAULT_SUBSTRATE_WS_PORT`,
```

test/bun.lock
ahmadkaouk, line 1: The changes here seem not related to the changes in the PR. We should revert them.

operator/Cargo.lock
ahmadkaouk, line 1: I don't see why the Cargo lock would have changed. It seems unrelated to the storagehub changes.

undercover-cactus: Ah it is probably when I was debugging the keys I had to patch the storage to add some more logs.

test/bun.lock
undercover-cactus: The changes have been reverted

test/launcher/storagehub-docker.ts
```diff
@@ -507,6 +452,63 @@ export const launchFishermanNode = async (
   logger.success(`Fisherman node started on port ${wsPort}`);
 };
 
+/**
+ * Launches a StorageHub Backend MSP container.
```
stiiifff: Naming is incorrect, should be `StorageHub MSP Backend`.

test/launcher/storagehub-docker.ts
```diff
@@ -507,6 +452,63 @@ export const launchFishermanNode = async (
...
+ *
+ * @param options - Configuration options for launching the network
+ * @param launchedNetwork - The launched network instance to track the node
+ */
+export const launchBackendMsp = async (
```
stiiifff: Fix naming (cfr. previous point).

test/launcher/storagehub-docker.ts
```diff
@@ -507,6 +452,63 @@ export const launchFishermanNode = async (
...
+  options: DataHavenOptions,
+  launchedNetwork: LaunchedNetwork
+): Promise<void> => {
+  logger.info("🚀 Launching StorageHub Backend MSP...");
+
+  const backendMSPImage = "moonsonglabs/storage-hub-msp-backend:latest";
+  const containerName = `storagehub-backend-msp-${options.networkId}`;
```
stiiifff: Fix naming (cfr. previous point).

test/launcher/storagehub-docker.ts
```diff
@@ -459,7 +409,7 @@ export const launchFishermanNode = async (
     `${wsPort}:${DEFAULT_SUBSTRATE_WS_PORT}`,
     options.datahavenImageTag,
     "--chain",
-    "dev",
+    "local",
```
stiiifff, line 412: The validators should also be explicitly started with the local (stagenet) chain preset.

test/launcher/storagehub-docker.ts
undercover-cactus: It is made on purpose so we don't get confuse with `StorageHub MSP`. Having `Backend` before `MSP` make it more obvious.

stiiifff: I find it actually even more confusing, `Backend MSP` is a term that's used nowhere else. Or you can drop MSP, and call it [StorageHub Backend](https://github.com/Moonsong-Labs/storage-hub/tree/main/backend).

PR #420, opened 2026-01-28
test: ✨ Add storage layout checks for upgradeable contracts

ahmadkaouk:
## Summary

  Implements storage layout testing for the upgradeable `DataHavenServiceManager` contract to prevent state
   corruption during proxy upgrades.

  ## Changes

  ### New Files
  - **`contracts/storage-snapshots/DataHavenServiceManager.storage.json`** - Baseline storage layout
  snapshot
  - **`contracts/storage-snapshots/README.md`** - Documentation for updating snapshots and known
  limitations
  - **`contracts/scripts/check-storage-layout.sh`** - CI script that compares current layout against
  snapshot
  - **`contracts/test/storage/StorageLayout.t.sol`** - Upgrade simulation tests verifying state
  preservation
  - **`.github/workflows/task-storage-layout.yml`** - CI workflow for storage layout checks

  ### Modified Files
  - **`.github/workflows/CI.yml`** - Added `storage-layout` job to run in parallel with other checks

  ## How It Works

  **Two-pronged approach:**

  1. **Snapshot Diff** - Compares current storage layout against committed snapshot using `forge inspect`.
  Catches unintended variable reordering, type changes, or gap modifications.

  2. **Upgrade Simulation** - Foundry tests that populate state, perform a proxy upgrade, and verify all
  values survive:
     - `test_upgradePreservesState` - Verifies core state variables
     - `test_upgradePreservesValidatorMappings` - Verifies `validatorEthAddressToSolochainAddress` mapping
     - `test_upgradePreservesMultipleValidators` - Verifies `validatorsAllowlist` with multiple entries
     - `test_functionalityAfterUpgrade` - Verifies contract remains functional post-upgrade

  ## Normalization

  The snapshot comparison normalizes JSON to avoid false positives:
  - Removes `astId` (changes with compiler runs)
  - Removes `contract` (contains full file path)
  - Removes `.types` section (contains unstable AST IDs embedded in type keys)
  - Sorts by slot number

  ## Usage

  ```bash
  # Check storage layout against snapshot
  ./scripts/check-storage-layout.sh

  # Run upgrade simulation tests
  forge test --match-contract StorageLayoutTest -vvv

  # Update snapshot (when intentionally changing storage)
  forge inspect DataHavenServiceManager storage --json >
  storage-snapshots/DataHavenServiceManager.storage.json
```
  ## Test Plan

  - ./scripts/check-storage-layout.sh passes
  - forge test --match-contract StorageLayoutTest -vvv passes (4 tests)
  - CI workflow runs successfully

### DIFF
--- a/.github/workflows/task-e2e.yml
+++ b/.github/workflows/task-e2e.yml
@@ -118,23 +118,6 @@ jobs:
           key: ${{ runner.os }}-bun-${{ hashFiles('**/bun.lock') }}
           restore-keys: |
             ${{ runner.os }}-bun-
-      - name: Cache Foundry libraries
-        uses: actions/cache/restore@v4
-        with:
-          path: ../contracts/lib
-          key: ${{ runner.os }}-foundry-libs-${{ hashFiles('.gitmodules') }}
-          restore-keys: |
-            ${{ runner.os }}-foundry-libs-
-
-      - name: Cache Foundry build artifacts
-        uses: actions/cache/restore@v4
-        with:
-          path: |
-            ../contracts/out
-            ../contracts/cache
-          key: ${{ runner.os }}-foundry-build-${{ hashFiles('contracts/foundry.toml', 'contracts/**/*.sol') }}
-          restore-keys: |
-            ${{ runner.os }}-foundry-build-
       - uses: docker/login-action@v3
         with:
           registry: ghcr.io
--- a/.github/workflows/task-foundry-tests.yml
+++ b/.github/workflows/task-foundry-tests.yml
@@ -38,16 +38,6 @@ jobs:
         with:
           version: v1.4.3
 
-      - name: Cache Foundry build artifacts
-        uses: actions/cache@v4
-        with:
-          path: |
-            contracts/out
-            contracts/cache
-          key: ${{ runner.os }}-foundry-build-${{ hashFiles('contracts/foundry.toml', 'contracts/**/*.sol') }}
-          restore-keys: |
-            ${{ runner.os }}-foundry-build-
-
       - run: forge --version
       - run: forge fmt --check
       - run: forge build --sizes

### CODE REVIEW

File: .github/workflows/CI.yml [hunk 1]
```diff
@@ -44,6 +44,8 @@ jobs:
     uses: ./.github/workflows/task-rust-tests.yml
   contract-tests:
     uses: ./.github/workflows/task-foundry-tests.yml
+  storage-layout:
+    uses: ./.github/workflows/task-storage-layout.yml
```
undercover-cactus, line 48: We should avoid adding new things in this file. Is there a particular reason to have it here ?

Out of the scope for this PR but I am wondering if the permissions set in this file are not then being set in the task called called via `uses`  (e.g ./.github/workflows/task-storage-layout.yml dont need write permission)

.github/workflows/task-storage-layout.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Cache Foundry build artifacts
+        uses: actions/cache@v4
+        with:
+          path: |
+            contracts/out
+            contracts/cache
+          key: ${{ runner.os }}-foundry-build-${{ hashFiles('contracts/foundry.toml', 'contracts/**/*.sol') }}
+          restore-keys: |
+            ${{ runner.os }}-foundry-build-
```
undercover-cactus: Is `forge build` that slow that we need cache ? Cache being scarce we should use it wisely.

stiiifff (commented):
@ahmadkaouk LGTM .. maybe adding a negative test, to prove the detection mechanism work, could be useful ?

.github/workflows/CI.yml
ahmadkaouk: I only added the storage-layout job in `CI.yml` to match the existing PR‑CI pattern we are using (all checks are wired up in this router and each calls a reusable. 

The called workflow’s permissions are what the caller grants (`CI.yml`); the reusable workflow can reduce them. In `task-storage-layout.yml` I explicitly set `permissions: contents: read`, so this job runs least‑privilege even though `CI.yml` grants `actions: write` / `packages: write` for other jobs.

.github/workflows/task-storage-layout.yml
ahmadkaouk: I don't think we need to cache forge build; it's very fast. The only reason I added it is because it was there before in the `forge test` job. I think we can get rid of it , but I'll probably do it in a follow up PR.

ahmadkaouk: In the end, I completely removed the cache because it was causing issues.

PR #421, opened 2026-01-29
fix: operator dockerfile to be used in CI and development

undercover-cactus:
## Summary

This PR remove the use of the `datahaven-dev.Dockerfile` for development and use the operator Dockerfile everywhere to make sure we are testing and developing using the same environment. 

## What changed

* Removed the `/data` volume
* Use of the `BINARY_PATH` build arg to specify the binary path
* Remove the builder extra step in the building process
* Added `--chain local` in our e2e tests scripts everywhere to make sure we are always using `local` in e2e tests/dev

### DIFF
--- a/.github/workflows/task-docker-ci.yml
+++ b/.github/workflows/task-docker-ci.yml
@@ -75,7 +75,7 @@ jobs:
           dockerfile: ./operator/Dockerfile
           context: .
           registry: ghcr.io
-          build_args: ./build/datahaven-node
+          build_args: BINARY_PATH=./build/datahaven-node
           registry_username: ${{ github.actor }}
           registry_password: ${{ secrets.GITHUB_TOKEN }}
           image_tags: ${{ steps.meta.outputs.tags }}


operator/Dockerfile
```diff
@@ -53,9 +53,9 @@ COPY --from=builder \
 
 # Create datahaven user and directories
 RUN useradd -m -u 1001 -U -s /bin/sh -d /datahaven datahaven && \
-    mkdir -p /datahaven/.local/share /data && \
-    chown -R datahaven:datahaven /data && \
-    ln -s /data /datahaven/.local/share/datahaven
+    mkdir -p /data /datahaven/.local/share /specs /storage && \
+    chown -R datahaven:datahaven /data /storage && \
```
ahmadkaouk: Should we also include `/specs` in the chown command, since it must be writable.

undercover-cactus: We actually don't need `/specs` for storagehub just `/storage`. I have fixed the description. The `/specs` was added in the dockerfile use locally but I am not sure what for.

operator/Dockerfile
```diff
@@ -53,9 +53,9 @@ COPY --from=builder \
...
+    ln -s /data /datahaven/.local/share/datahaven-node
```
stiiifff: The symlink is not for the binary but for a datahaven folder.

operator/Dockerfile
```diff
@@ -53,9 +53,9 @@ COPY --from=builder \
 
 # Create datahaven user and directories
 RUN useradd -m -u 1001 -U -s /bin/sh -d /datahaven datahaven && \
-    mkdir -p /datahaven/.local/share /data && \
-    chown -R datahaven:datahaven /data && \
-    ln -s /data /datahaven/.local/share/datahaven
+    mkdir -p /data /datahaven/.local/share /specs /storage && \
```
stiiifff: These additional folders don't bring anything .. you could just map several subfolders under the data folder.

stiiifff (changes requested):

operator/Dockerfile
undercover-cactus: Those changes are from the Dockerfile used to run the nodes locally. I have added it so what is running locally is also running in the CI : https://github.com/datahaven-xyz/datahaven/blob/main/docker/datahaven-dev.Dockerfile

PR #422, opened 2026-01-29
feat: Add DH-AVS stagenet/testnet Hoodi deployment support

stiiifff:
## Summary

- Add multi-environment deployment support (stagenet, testnet, mainnet) to CLI and contracts
- Configure stagenet and testnet runtimes with correct genesis hashes and Snowbridge Agent IDs
- Add CLI commands for BEEFY checkpoint updates and rewards origin computation
- Add ETH validator strategies (native beacon chain ETH + LSTs) to all config files

### Runtime Configuration

**Stagenet Runtime:**
- Set `StagenetGenesisHash` to DataHaven stagenet genesis hash
- Configure `RewardsAgentOrigin` with computed Snowbridge Agent ID
- Add tests verifying rewards account derivation and agent ID computation

**Testnet Runtime:**
- Set `TestnetGenesisHash` to DataHaven testnet genesis hash
- Configure `RewardsAgentOrigin` with computed Snowbridge Agent ID
- Add tests verifying rewards account derivation and agent ID computation

The Rewards Agent ID is computed following Snowbridge's location description pattern:
```
blake2_256(SCALE_ENCODE("GlobalConsensus", ByGenesis(genesis), "AccountKey20", rewards_account))
```

### CLI Enhancements

- All contracts subcommands (`status`, `deploy`, `verify`, `update-metadata`) now accept `--environment` option
- Config and deployment files use environment-prefixed naming (e.g., `stagenet-hoodi.json`, `testnet-hoodi.json`)
- New `update-beefy-checkpoint` command that:
  - Connects to a live DataHaven chain via WebSocket RPC
  - Fetches all BEEFY data at the same finalized block for consistency
  - Uses parallel queries with `Promise.all` for better performance
  - Computes authority hashes (keccak256 of Ethereum addresses derived from BEEFY public keys)
  - Uses Snowbridge's quorum formula `n - floor((n-1)/3)` for strictly > 2/3 majority
- New `update-rewards-origin` command that computes the Snowbridge Agent ID for the rewards pallet
- Centralized validation via `contractsPreActionHook` for all contract commands
- Environment validation against allowlist (`stagenet`, `testnet`, `mainnet`)

### Contract Changes

- Network validation uses explicit allowlist instead of suffix matching
- Added `initialValidatorSetId` and `nextValidatorSetId` fields to `SnowbridgeConfig` struct
- `DeployBase.s.sol` now uses config values for validator set IDs instead of hardcoded 0/1
- `DeployParams.s.sol` loads validator set IDs from config with backwards compatibility

### Validator Strategies

Added ETH-equivalent strategies to allow validators to stake using native ETH or LSTs:

**All Networks:**
- `0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0` - Native beacon chain ETH (virtual strategy)

**Hoodi Testnet:**
- `0xf8a1a66130d614c7360e868576d5e59203475fe0` - stETH
- `0x24579aD4fe83aC53546E5c2D3dF5F85D6383420d` - WETH

**Ethereum Mainnet:**
- `0x93c4b944D05dfe6df7645A86cd2206016c51564D` - stETH
- `0x1BeE69b7dFFfA4E2d53C2a2Df135C388AD25dCD2` - rETH
- `0x54945180dB7943c0ed0FEE7EdaB2Bd24620256bc` - cbETH

### Config Files

- `stagenet-hoodi.json` - Hoodi testnet with stagenet EigenLayer addresses
- `testnet-hoodi.json` - Hoodi testnet with testnet EigenLayer addresses  
- `mainnet-ethereum.json` - Ethereum mainnet with mainnet EigenLayer addresses
- Removed `hoodi.json` (replaced by environment-prefixed files)

## Usage

```bash
# Deploy to stagenet on Hoodi
bun cli contracts deploy --chain hoodi --environment stagenet

# Update BEEFY checkpoint from live chain
bun cli contracts update-beefy-checkpoint \
  --chain hoodi \
  --environment stagenet \
  --rpc-url wss://services.datahaven-dev.network/stagenet

# Compute rewards origin for a chain
bun cli contracts update-rewards-origin \
  --chain hoodi \
  --environment stagenet \
  --rpc-url wss://services.datahaven-dev.network/stagenet

# Check deployment status
bun cli contracts status --chain hoodi --environment stagenet
```


contracts/config/hoodi.json
ahmadkaouk, line 1: Do we still need this file? Is it used anywhere? From what I understand, `stagenet-hoodi.json` and `testnet-hoodi.json` files replace it.

contracts/script/deploy/DeployLive.s.sol
```diff
@@ -209,23 +210,58 @@ contract DeployTestnet is DeployBase {
...
+
+        uint256 offset = networkBytes.length - chainBytes.length;
+        for (uint256 i = 0; i < chainBytes.length; i++) {
+            if (networkBytes[offset + i] != chainBytes[i]) {
+                return false;
+            }
+        }
+
+        return true;
     }
```
ahmadkaouk: The problem with this approach to validate the network name is that it will return true for any string ending with `hoodi` or `ethereum` like `abcdhoodi`. I believe a simple allowlist is cleaner and more explicit:
```
function _validateNetwork(string memory network) internal pure {
    bytes32 h = keccak256(bytes(network));
    
    if (
        h == keccak256("hoodi") ||
        h == keccak256("stagenet-hoodi") ||
        h == keccak256("testnet-hoodi") ||
        h == keccak256("mainnet-ethereum") ||
        h == keccak256("ethereum")
    ) {
        return;
    }

    revert(
        string.concat(
            "Unsupported network: ",
            network,
            ". Supported: hoodi, stagenet-hoodi, testnet-hoodi, ethereum, mainnet-ethereum"
        )
    );
}
```

test/cli/handlers/contracts/beefy-checkpoint.ts
```diff
@@ -0,0 +1,327 @@
...
+  const compressedKeyHex = compressedPubKey.startsWith("0x")
+    ? compressedPubKey.substring(2)
+    : compressedPubKey;
+
+  const point = secp256k1.ProjectivePoint.fromHex(compressedKeyHex);
+  const uncompressedPubKeyBytes = point.toRawBytes(false);
+  const uncompressedPubKeyHex = toHex(uncompressedPubKeyBytes);
+
+  return publicKeyToAddress(uncompressedPubKeyHex);
+};
```
ahmadkaouk: The code is duplicated. The same function is already defined in `datahaven.ts`. we can import it from there.

test/cli/handlers/contracts/beefy-checkpoint.ts
```diff
@@ -0,0 +1,327 @@
...
+ * Uses ceil(n * 2/3) where n is the number of validators.
+ *
+ * @param validatorCount - The number of validators
+ * @returns The minimum number of required signatures
+ */
+const calculateMinRequiredSignatures = (validatorCount: number): number => {
+  // For BFT security, we need > 2/3 of validators to sign
+  // Using ceil(n * 2/3) ensures we have more than 2/3 threshold
+  return Math.ceil((validatorCount * 2) / 3);
+};
```
ahmadkaouk, line 65: The function iss using `Math.ceil((n * 2) / 3)` which returns exactly 2/3 when the validator count is a multiple of 3 (e.g., n=3 returns 2, n=6 returns 4). BEEFY consensus requires strictly more than 2/3. We should use the same formula used in the Snowbridge BeefyClient contract https://github.com/datahaven-xyz/snowbridge/blob/13263fefa29a3f4af50e5650dcd93fe3afac44db/contracts/src/BeefyClient.sol#L488-L496.

test/cli/handlers/contracts/beefy-checkpoint.ts
```diff
@@ -0,0 +1,327 @@
...
+    const minNumRequiredSignatures = calculateMinRequiredSignatures(currentAuthorityKeys.length);
+    logger.info(
+      `📊 Minimum required signatures: ${minNumRequiredSignatures} (ceil(${currentAuthorityKeys.length} * 2/3))`
+    );
+
+    // Fetch next authorities (NextAuthorities)
+    logger.info("🔍 Fetching BEEFY NextAuthorities (next set)...");
+    const nextAuthoritiesRaw = await dhApi.query.Beefy.NextAuthorities.getValue({
+      at: "best"
+    });
```
ahmadkaouk: We are querying `startBlock` from `getFinalizedBlock()` but we fetch ValidatorSetId, Authorities, NextAuthorities at at: "best". This can produce a mismatch. We should query the authorities from the same `startBlock`. Also, we could benefit from reading all the values in parallel using `promise.all` and not sequentially .

test/cli/handlers/contracts/deploy.ts
```diff
@@ -49,12 +62,21 @@ export const contractsCheck = async (options: any, command: any) => {
...
+  // Get environment option
+  let environment = options.environment;
+  if (!environment && command.parent) {
+    environment = command.parent.getOptionValue("environment");
+  }
+
+  // Build network identifier with environment prefix if specified
+  const networkId = environment ? `${environment}-${chain}` : chain;
+
+  printHeader(`Checking DataHaven ${networkId} Configuration and Status`);
```
ahmadkaouk: This pattern is repeated multiple times in the file and is redundant. We could extract it into a helper.

test/cli/index.ts
```diff
@@ -199,15 +201,20 @@ const contractsCommand = program
...
+    - update-rewards-origin: Fetch or compute the RewardsAgentOrigin and update config
     - update-metadata: Update the metadata URI of an existing AVS contract
-    
+
     Common options:
-    --chain: Target chain (required: hoodi, mainnet, anvil)
+    --chain: Target chain (required: hoodi, ethereum, anvil)
+    --environment: Deployment environment (stagenet, testnet, mainnet)
+                   When specified, config files are read from {environment}-{chain}.json
+                   and deployments are written to {environment}-{chain}.json
```
ahmadkaouk, line 217: I find the current pattern of constructing `networkId` from `--environment` + `--chain` a bit complicated and error-prone. Since `--environment` is optional and its value isn’t validated, it’s easy to end up with a `networkId` that passes the `chain` checks but fails later at runtime when we try to read/write config files.

why not simply using a single explicit flag (e.g. `--network stagenet-hoodi | testnet-hoodi | mainnet-ethereum | anvil`) and validate it against a known allowlist? This would make the UX clearer and fail fast with a helpful error message.

gonzamontiel (changes requested):
I would be explicit with the combinations supported for --environment.
Basically, implicitly we are saying something like:
- `stagenet` / `testnet` → should only work with `hoodi` (testnet)
- `mainnet` → should only work with `ethereum` (mainnet)
- `anvil` → local only, no need accept environment at all

It would be interesting to have some safeguard like:
```
const VALID_COMBINATIONS: Record<string, string[]> = {
  hoodi: ["stagenet", "testnet"], 
  ethereum: ["mainnet"],  
  anvil: [], 
};

function validateChainEnvironment(chain: string, environment?: string) {...}
```

So we don't assume that the created configs are the only valid ones, and we forbid mixing a mainnet with a testnet.

test/cli/index.ts
```diff
@@ -289,25 +358,47 @@ contractsCommand
...
+    }
+    const supportedChains = ["hoodi", "ethereum", "anvil"];
+    if (!supportedChains.includes(chain)) {
+      console.error(
+        `❌ Unsupported chain: ${chain}. Supported chains: ${supportedChains.join(", ")}`
+      );
+      process.exit(1);
+    }
+    await contractsCheck(options, command);
+  });
```
ahmadkaouk, line 391: What's the reason for removing the preAction hook and duplicating the chain validation ?

contracts/config/hoodi.json
stiiifff: Removed in c16a830af63301f312fda8d5bbe9c48888724ea0

contracts/script/deploy/DeployLive.s.sol
stiiifff: Fixed in c16a830af63301f312fda8d5bbe9c48888724ea0

test/cli/handlers/contracts/beefy-checkpoint.ts
stiiifff: Good catch. Fixed in c16a830af63301f312fda8d5bbe9c48888724ea0

test/cli/handlers/contracts/beefy-checkpoint.ts
stiiifff: Fixed in c16a830af63301f312fda8d5bbe9c48888724ea0

test/cli/handlers/contracts/beefy-checkpoint.ts
stiiifff: Fixed in c16a830af63301f312fda8d5bbe9c48888724ea0

test/cli/handlers/contracts/deploy.ts
stiiifff: Fixed in c16a830af63301f312fda8d5bbe9c48888724ea0

test/cli/index.ts
stiiifff: I think it's still probably useful to be able to deploy (or other command) contracts to a specific chain (local, test or prod) without specifying an environment. But I don't have a strong opinion on this.

test/cli/index.ts
stiiifff: Restored in c16a830af63301f312fda8d5bbe9c48888724ea0

PR #428, opened 2026-02-03
fix: Register the snowbridge agent in the Dathaven Service instead of the operator node 

undercover-cactus:
## Summary

This PR rename the `rewardsAgentOrigin`, `rewardsMessageOrigin`, etc... into a less specific less now that the Snowbrige Agent is also being used to relay slashing messages.

This PR also have a fix to register the Agent address instead of the operator node address to check the sender of the message. Without this fix we could never relay rewards or execute slashing because we would get an error regarding the message.

## What changed 

* Removing the prefix `rewards` everytime we were refering the snowbridge agent (to clarify that the agent is not only being used by the reward feature)
* Fix the deployment script to register the `agentAddress` as the required sender for relaying substrate message

## What is missing

[ ] ~~Rename `onlyRewardsInitiator` and `rewardsInitiator` in the `DatahavenServiceManager.sol ` for something that would include slashing~~ This should be done in another PR.
[x] Check the Testnet Deploy script to  make sure it is using the agent address

### DIFF
--- a/test/.dockerignore
+++ b/test/.dockerignore
@@ -1,0 +1,15 @@
+# Keep submitter image build context minimal.
+*
+
+!package.json
+!bun.lock
+!tsconfig.json
+!bunfig.toml
+!.papi/
+!.papi/**
+!tools/validator-set-submitter/
+!tools/validator-set-submitter/**
+!contract-bindings/
+!contract-bindings/**
+!utils/
+!utils/**
--- a/test/e2e/framework/submitter.ts
+++ b/test/e2e/framework/submitter.ts
@@ -17,13 +17,13 @@ const SUBMITTER_READY_TIMEOUT_SECONDS = 30;
 const SUBMITTER_LOG_TAIL_LINES = 200;
 
 /**
- * Builds the validator-set-submitter Docker image from the repo root.
+ * Builds the validator-set-submitter Docker image from the test directory.
  */
 export async function buildSubmitterImage(): Promise<void> {
   logger.debug("Building validator-set-submitter Docker image...");
-  const repoRoot = path.resolve(import.meta.dir, "../../..");
-  await $`docker build -f test/tools/validator-set-submitter/Dockerfile -t ${SUBMITTER_IMAGE} .`
-    .cwd(repoRoot)
+  const testRoot = path.resolve(import.meta.dir, "../..");
+  await $`docker build -f tools/validator-set-submitter/Dockerfile -t ${SUBMITTER_IMAGE} .`
+    .cwd(testRoot)
     .quiet();
   logger.debug("Validator-set-submitter image built successfully");
 }
@@ -106,9 +106,11 @@ export async function launchSubmitter(options: LaunchSubmitterOptions): Promise<
       timeoutSeconds: SUBMITTER_READY_TIMEOUT_SECONDS
     });
   } catch (error) {
+    const logResult = await $`docker logs --tail ${SUBMITTER_LOG_TAIL_LINES} ${containerName}`
+      .nothrow()
+      .quiet();
     const logs =
-      (await $`docker logs --tail ${SUBMITTER_LOG_TAIL_LINES} ${containerName}`.nothrow().text()) ||
-      "<no logs captured>";
+      `${logResult.stdout.toString()}${logResult.stderr.toString()}`.trim() || "<no logs captured>";
     await stopSubmitter(containerName);
     throw new Error(
       `Submitter did not become ready. Expected log "${SUBMITTER_READY_LOG}". Last ${SUBMITTER_LOG_TAIL_LINES} log lines:\n${logs}`,
--- a/test/e2e/framework/suite.ts
+++ b/test/e2e/framework/suite.ts
@@ -3,6 +3,7 @@ import readline from "node:readline";
 import { isCI } from "launcher/network";
 import { logger } from "utils";
 import { launchNetwork } from "../../launcher";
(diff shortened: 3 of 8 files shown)


operator/runtime/stagenet/src/configs/runtime_params.rs
```diff
@@ -51,13 +51,13 @@ pub mod dynamic_params {
...
         #[allow(non_upper_case_globals)]
-        /// The RewardsAgentOrigin is the Agent ID for the rewards pallet's outbound Snowbridge messages.
+        /// The AgentOrigin is the Agent ID for the rewards/slashes pallet's outbound Snowbridge messages.
         /// Computed as: blake2_256(SCALE_ENCODE("GlobalConsensus", ByGenesis(genesis_hash), interior))
         /// where interior = SCALE_ENCODE("AccountKey20", ExternalValidatorRewardsAccount)
         ///
         /// For stagenet with genesis hash 0x72d0856fd339e09cb21df7bac8ac3120bd871e327ec0e1658395df68acef9bee
         /// and rewards account 0x6d6f646c64682f65767265770000000000000000 (from PalletId "dh/evrew"):
-        pub static RewardsAgentOrigin: H256 = H256::from_slice(&hex!(
+        pub static AgentOrigin: H256 = H256::from_slice(&hex!(
```
stiiifff, line 60: What's the impact of renaming an existing dynamic runtime parameter ? (i.e. will the runtime upgrade on a live network work without any issue ?)

operator/runtime/testnet/src/configs/runtime_params.rs
```diff
@@ -49,13 +49,13 @@ pub mod dynamic_params {
...
         #[allow(non_upper_case_globals)]
-        /// The RewardsAgentOrigin is the Agent ID for the rewards pallet's outbound Snowbridge messages.
+        /// The AgentOrigin is the Agent ID for the rewards/slashes pallet's outbound Snowbridge messages.
         /// Computed as: blake2_256(SCALE_ENCODE("GlobalConsensus", ByGenesis(genesis_hash), interior))
         /// where interior = SCALE_ENCODE("AccountKey20", ExternalValidatorRewardsAccount)
         ///
         /// For testnet with genesis hash 0xdbf403d348916fb0694485bc7f9c0d8c53fdf86664ebac019af209c090c3df99
         /// and rewards account 0x6d6f646c64682f65767265770000000000000000 (from PalletId "dh/evrew"):
-        pub static RewardsAgentOrigin: H256 = H256::from_slice(&hex!(
+        pub static AgentOrigin: H256 = H256::from_slice(&hex!(
```
stiiifff, line 58: Same question as for the Stagenet runtime.

operator/runtime/stagenet/src/configs/runtime_params.rs
undercover-cactus: As far as I can tell it should not have an impact to change the variable name because what is looking at is the `index = 3` and not the name of the variable in Rust. The name variable only has an impact at the Rust level.

We didn't change the index so the slot of the storage should be the same.

However I cannot find anything in the documentation explicit about it.

operator/runtime/testnet/src/configs/runtime_params.rs
undercover-cactus: Same as before. Because we kept the `index = 3` value it should not have an impact.

operator/runtime/stagenet/src/configs/runtime_params.rs
gonzamontiel: I agree, changing the type or the index would be breaking, changing the rust identifier shouldn't be an issue.

contracts/script/deploy/DeployParams.s.sol
```diff
@@ -24,8 +24,7 @@ contract DeployParams is Script, Config {
         config.minNumRequiredSignatures =
             vm.parseJsonUint(configJson, ".snowbridge.minNumRequiredSignatures");
         config.startBlock = vm.parseJsonUint(configJson, ".snowbridge.startBlock").toUint64();
-        config.rewardsMessageOrigin =
-            vm.parseJsonBytes32(configJson, ".snowbridge.rewardsMessageOrigin");
+        config.messageOrigin = vm.parseJsonBytes32(configJson, ".snowbridge.messageOrigin");
```
ahmadkaouk, line 27: We now read `.snowbridge.messageOrigin` but `mainnet-ethereum.json` and `testnet-hoodi.json` still have `rewardsMessageOrigin`

undercover-cactus: Everything should be changed now. There was some conflicts with a previous PR that has been merged.

operator/runtime/testnet/src/configs/mod.rs
```diff
@@ -1669,7 +1669,7 @@ impl datahaven_runtime_common::slashes_adapter::SlashesSubmissionConfig for Test
     }
 
     fn slashes_agent_origin() -> H256 {
-        runtime_params::dynamic_params::runtime_config::RewardsAgentOrigin::get()
+        runtime_params::dynamic_params::runtime_config::AgentOrigin::get()
         // TODO: Can we use the same as reward and just rename the config to `AgentOrigin` ?
```
ahmadkaouk:
```suggestion
```
This todo is resolved now

contracts/deployments/anvil-rewards-info.json
ahmadkaouk, line 1: I wonder if this file is still useful or should be removed since deployments now  will output `{network}-agent-info.json`

PR #454, opened 2026-02-23
fix: verify allowlist tx receipt status in e2e validator helper

vikions:
## Summary

`addValidatorToAllowlist` in the E2E validator helper was awaiting 
`waitForTransactionReceipt` but discarding the result entirely. 
This means a reverted transaction would still be treated as a 
successful allowlist operation — the test would continue without 
any indication that the on-chain call failed.

This patch captures the receipt and explicitly checks its status, 
throwing a descriptive error if the transaction did not succeed. 
This prevents false-positive test results when the allowlist tx reverts.

## Changes
- `test/e2e/framework/validators.ts`: capture receipt from 
  `waitForTransactionReceipt` and assert `status === "success"`

## Test plan
- Existing E2E validator test suites exercise this code path
- A reverted allowlist tx will now surface as an explicit failure 
  rather than passing silently


stiiifff:
Thanks for your contribution @vikions !

PR #472, opened 2026-03-11
fix: include PoV base cost estimation for `EthereumRuntimeAPI::create()`

gonzamontiel:
## Summary
In the `EthereumRuntimeAPI::create()` API, the runtime now passes a proof-size base cost into Frontier’s runner so that PoV validation accounts for the encoded transaction size, not only the gas-derived weight. Before the parameter was set to `None`.

- Estimated transaction length is computed for an EIP1559-style create: base 190 bytes (from, value, gas_limit, nonce, action, chain_id, signature) plus data.len(), optional max_fee_per_gas / max_priority_fee_per_gas (32 bytes each when present), and encoded access_list length.
- Conditional passing: when the converted weight has proof_size() > 0, the runtime passes (weight_limit, proof_size_base_cost) into Runner::create(); otherwise it passes (None, None) so behavior is unchanged when proof size is not used.
- Applied change in mainnet, stagenet, and testnet runtimes (operator/runtime/*/src/lib.rs)


ahmadkaouk (commented):
Can you clarify the concrete problem this fixes for DataHaven?
 
DataHaven is a solo chain, and in our runtimes we explicitly do not account for PoV today (`GasLimitPovSizeRatio = 0`). So as far as I can tell, `gas_to_weight(...).proof_size()` stays `0`, and this new branch still resolves to `(None, None)`, meaning it does not change behavior for us currently. I’m not convinced this should go in as-is.

stiiifff:
**Blocking issues**

  1. The new branch never fires — proof_size() is always 0.
  - All three runtimes set `GasLimitPovSizeRatio = 0` `(operator/runtime/{mainnet,stagenet,testnet}/src/configs/mod.rs:1024/1021/1024)`.
  - `WeightPerGas = Weight::from_parts(WEIGHT_PER_GAS, 0)` — `proof_size` component is `0` per gas unit
  `(configs/mod.rs:1019/1016/1019)`.
  - So `gas_to_weight(gas, true).proof_size()` is always `0`, and the new conditional resolves to `(None, None)` in 100% of
  cases. The encoded-length arithmetic is computed and discarded.

  2. Subtle behavior regression despite the "no-op" appearance.
  - Before: `Some(weight_limit), None` → runner enforced a `ref_time `budget derived from gas_limit.
  - After: `None, None` → runner runs with no weight limit.
  - This likely explains the gas estimate drop in test-gas-contract-creation.ts (`210541` → `156082`) and
  `test-gas-estimation-contracts.ts` (`255341` → `172902`). These changes are a side effect of removing the `ref_time` limit,
  not of the PoV accounting the PR title advertises. Worth confirming this is intentional — if so, the PR description
  and title should be rewritten accordingly.

  3. Asymmetric fix: `call()` is not updated.
  - `fn call()` at `mainnet/src/lib.rs:1094` still passes `Some(weight_limit), None` with the same issue the PR claims to fix.
   Frontier's upstream template applies PoV base cost to both call and create.

gonzamontiel:
I would say we can close this one, there's no big reason to continue working on this / fixing the misbehavior mentioned above.

PR #478, opened 2026-03-18
fix: deregister operator on removal from allowlist

gonzamontiel:
## Summary
This PR makes `removeValidatorFromAllowlist` immediately revoke active validator membership instead of only preventing future registrations. Previously, removing a validator from the allowlist only blocked future registration attempts, but did not remove validators that were already active in the operator set. This change closes that gap so allowlist removal immediately takes effect for active validator membership as well.

### What changed
- Updated `DataHavenServiceManager.removeValidatorFromAllowlist` to force-deregister validators from the `VALIDATORS_SET_ID` when they are currently registered.
- Extracted the deregistration flow into an internal helper so the same deregistration logic is reused consistently.

### Tests
Added regression tests covering:
  - removing an allowlisted validator that is actively registered
  - removing an allowlisted validator that was never registered
  - ensuring removed validators are excluded from validator-set message generation


ahmadkaouk (commented):
`removeValidatorFromAllowlist` now force-deregisters the validator, which ends up calling deregisterOperator and clearing `validatorSolochainAddressToEthAddress` immediately. The problem is that EigenLayer still keeps a deregistered operator slashable until `DEALLOCATION_DELAY`, while `slashValidatorsOperator` can only resolve the ETH operator through that mapping. So during the post-deregistration slashable window, a slash by solochain address is silently skipped even though `AllocationManager` still considers the operator slashable.

 This same change also alters rewards behavior: `submitRewards` now drops any late/historical rewards for a just-removed validator as an "unknown" solochain address, because the mapping has already been deleted.

ahmadkaouk (approved):

stiiifff (approved):

